Issue #11, opened 2018-07-27
ff-communication API documentation - Currency Guide - Broken Link


sebastiantri:
The link to the currency guide within the API documentation of ff-communication are broken:
Is:
https://web-components.fact-finder.de/guides/currency-guide 
Should: 
https://web-components.fact-finder.de/documentation/currency-guide

damrath:
In Review [#13]

Issue #10, opened 2018-07-27
ff-communication API documentation


sebastiantri:
Some properties in the API documentation are wrongly described (boolean with two options false, false).
E.g. 
disable-single-hit-redirect (String) Options:  **false,  false** (default: false)

Issue #9, opened 2018-07-24
factfinder.communication.ResultDispatcher callback second event param doc is missing


toarm:
the event is now passed as a second param in ResultDispatcher.subscribe callback

Issue #8, opened 2018-07-18
factfinder.communication.ResultDispatcher.subscribe("navigation", callback)


toarm:
ResultDispatcher doc for navigation topic is missing

PR #13, opened 2018-07-27
Issues: 8 - 11 | Fix errors in documentation

damrath:
Links updated, updated ResultDispatcher callback documentation and updated some erroneous parameters.

#8 #9 #10 #11


.gitignore
```diff
@@ -2,3 +2,4 @@ build/
 node_modules/
 .idea
 *.iml
+package-lock.json
```
KrzysztofKozlowskiOBSS: I guess we should commit this file

damrath: i don't think we need it though. What do you say @toarm ?

KrzysztofKozlowskiOBSS: https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5

damrath: I know they recommend doing so, however we haven't been doing this consistently, neither have we used the feature afaik. @toarm should decide

toarm: Honestly I never thought about package-lock.json. For me it feels initially wrong to push such a file but on the other side they are recommending it x/

I'd like to test it. If it results in merge conflicts and other annoying stuff we can remove it. Acutally I dont see any benefit in pushing it but lets stick to the recommendations for a while and see if it works out...

damrath: @toarm @KrzysztofKozlowskiOBSS re-added package-lock.json

PR #36, opened 2018-08-16
Fix broken EventAggregator link

KrzysztofKozlowskiOBSS:


KrzysztofKozlowskiOBSS:
Should be branched from dev.

PR #37, opened 2018-08-16
Fix broken EventAggregator link

KrzysztofKozlowskiOBSS:


markdown/en/core-result-dispatcher.md
```diff
@@ -4,7 +4,7 @@ subscribe to ResultDispatcher to get notified when new data is available.
 
 The ResultDispatcher is in the module: **factfinder.communication.ResultDispatcher**.
 
-If you want to trigger your own communication events checkout the documentation for the [FFCommunicationEventAggregator](documentation/core-event-aggregator).
+If you want to trigger your own communication events checkout the documentation for the [FFCommunicationEventAggregator](api/core-event-aggregator).
```
omikron-fp: I would make it `api/core-event-aggregator#tab=docs` to the link. When the fragment is added by the app, another entry in the browser history is created and you have to click twice to navigate back. (At least in my Chrome)

KrzysztofKozlowskiOBSS:
@omikron-fp can you check if we can remove the history entry without hash param with #tab=docs instead of creating a new one?

omikron-fp:
@KrzysztofKozlowskiOBSS Sure, that would actually be the more desirable behaviour. Might take me a little to dive into the code.

KrzysztofKozlowskiOBSS:
@omikron-fp https://github.com/FACT-Finder-Web-Components/documentation/blob/2234b30df9c6e236676c7c8cb0c26ddae30a4fd6/src/views/api-view.js#L198-L199 if you manage to replace the location hash here manually, our router shouldn't add a new history entry

PR #43, opened 2018-08-22
FFWEB-727 documentation 404 page not shown

omikron-fp:
This fix adds URL validation and displays the 404 page if necessary.
It is good enough to be deployed.

A more sophisticated 404 page is in progress and will be provided later.


src/views/view-404.js
```diff
@@ -12,15 +12,30 @@ import {PolymerElement, html} from '@polymer/polymer/polymer-element.js';
...
+        width: 100%;
+        max-width: 800px;
+        border-radius: 50%;
+    }
 </style>
 
-Oops you hit a 404. <a href="[[rootPath]]">Head back to home.</a>
+<h2>Oops! You hit a 404!</h2>
+<div class="imgContainer">
+    <img src="http://search-web-components.fact-finder.de/crap/img/stupidCat.jpg">
```
damrath: You should download the asset.

toarm: yeah, its located in another project and as soon as that dissapears its gone in our documentation x/

src/data/api.js
```diff
@@ -107,8 +107,8 @@ const api = {
         },
         "ff-template": {
             path: "ff-template",
-            noApi: true
             title: "Template",
+            noApi: true  // TODO find out what this is for and remove if not needed
```
toarm: as far a i can remember its for pages without an API tab

omikron-fp: `noApi` never gets referenced anywhere. The reason `Template`'s API tab shows a placeholder message is because `api-view.js` handles the failed markdown load itself in the `connectedCallback`.

I think we can safely remove it. At the moment it doesn't do anything anyway. Even giving other pages the `noApi: true` property has no effect at all.

PR #50, opened 2018-08-28
use relative link

### LINKED ISSUE #49: Use relative URL in record-list
FACT-Finder-Web-Components/documentation-export#2

Export will use relative URLs.
The record-list needs to be changed so it opens relative instead of absolute links.

damrath:
fixes #49

### DIFF
--- a/src/search/sd-record-list.js
+++ b/src/search/sd-record-list.js
@@ -90,7 +90,7 @@ class SdRecordList extends GestureEventListeners(PolymerElement) {
     goToDeeplink(e) {
         console.log(e);
         if (e.model.__data.item.record.deeplink !== "") {
-            window.open(e.model.__data.item.record.deeplink, "_blank")
+            window.open(window.location.domain + e.model.__data.item.record.deeplink, "_blank")
         }
     }
 
--- a/src/search/sd-record-list.js
+++ b/src/search/sd-record-list.js
@@ -88,11 +88,13 @@ class SdRecordList extends GestureEventListeners(PolymerElement) {
     }
 
     goToDeeplink(e) {
-        console.log(e);
-        if (e.model.__data.item.record.deeplink !== "") {
-            window.open(window.location.domain + e.model.__data.item.record.deeplink, "_blank")
-        }
+        const deeplink = e.model.__data.item.record.deeplink;
+        if (!deeplink) return;
+
+        const targetLink = deeplink.startsWith("http") ? deeplink : window.location.origin + deeplink;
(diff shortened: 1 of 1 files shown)


src/search/sd-record-list.js
```diff
@@ -90,7 +90,7 @@ class SdRecordList extends GestureEventListeners(PolymerElement) {
     goToDeeplink(e) {
         console.log(e);
         if (e.model.__data.item.record.deeplink !== "") {
-            window.open(e.model.__data.item.record.deeplink, "_blank")
+            window.open(window.location.domain + e.model.__data.item.record.deeplink, "_blank")
```
omikron-fp: Can't reproduce the fix on local machine.
Opens a URL like this: `undefinedhttps://web-components.fact-finder.de/api/ff-template#tab=docs`.
`e.model.__data.item.record.deeplink` returns an absolute URL.
What is `domain` supposed to be? `origin` perhaps?
```
window.location = {
  "href": "http://localhost:8081/search?query=asn&channel=webc-doku-text&sid=...",
  "ancestorOrigins":{},
  "origin": "http://localhost:8081",
  "protocol": "http:",
  "host": "localhost:8081",
  "hostname": "localhost",
  "port": "8081",
  "pathname": "/search",
  "search": "?query=asn&channel=webc-doku-text&sid=...",
  "hash": ""
}
```

damrath: You're right, domain isn't always defined. `host` would be more suitable

src/search/sd-record-list.js
```diff
@@ -88,9 +88,13 @@ class SdRecordList extends GestureEventListeners(PolymerElement) {
...
-        console.log(e);
-        if (e.model.__data.item.record.deeplink !== "") {
-            window.open(window.location.origin + e.model.__data.item.record.deeplink, "_blank")
+        let deeplink = e.model.__data.item.record.deeplink;
+        if (deeplink) {
+            if (deeplink.startsWith("http")) {
+                window.open(deeplink, "_blank")
+            } else {
+                window.open(window.location.origin + deeplink, "_blank")
+            }
```
omikron-fp: `deeplink` doesn't change. No need to allow it to do so.
We can go without nested branching here and with a single action statement.
What about:
```
const deeplink = e.model.__data.item.record.deeplink;
if (!deeplink) return;

const targetLink = deeplink.startsWith("http") ? deeplink : window.location.origin + deeplink;
window.open(targetLink, "_blank");
```

PR #55, opened 2018-08-31
[#38] don't show api tab if "noAPi" property is true

### LINKED ISSUE #38: ff-template API documentation is missing.
The API tab of the ff-template is broken 
https://web-components.fact-finder.de/api/ff-template#tab=api

damrath:
fixes #38 

There is a `noApi` property in `api.js` which apparently was intended to prevent the API tab from being shown on the documentation page, but it was never checked, hence #38 

I added a check to fix this issue, now `noApi: true` will prevent the rendering of an api tab.

### DIFF
--- a/src/views/api-view.js
+++ b/src/views/api-view.js
@@ -108,7 +108,9 @@ class ApiView extends ViewMixin(ReduxMixin(PolymerElement)) {
                 <paper-tabs selected="{{activeTab}}" attr-for-selected="name">
                     <paper-tab name="docs">Documentation</paper-tab>
                     <paper-tab name="demo">Demo</paper-tab>
+                    <template is="dom-if" if="{{api}}">
                     <paper-tab name="api">Api</paper-tab>
+                    </template>
                 </paper-tabs>
             </app-toolbar>
         </template>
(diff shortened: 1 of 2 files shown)


src/views/api-view.js
```diff
@@ -164,10 +166,18 @@ class ApiView extends ViewMixin(ReduxMixin(PolymerElement)) {
...
+                type: Boolean,
+                computed: 'computeApi(subpage, data)'
             }
         }
     }
 
+    computeApi(pageName, data) {
+        if(pageName && data) return !data.pages[pageName].noApi;
+    }
+
```
omikron-fp: 1)
You have two return types here. `bool` and well.... an implicit `undefined`.
I think it would be more obvious to see the purpose of this function's result if it used explicit `return`s for all possibilities.
I'd go for something like `return pageName && data && !data.pages[pageName].noApi` so the return type is always a `bool`.

2)
Plus, I think the function's name could be more expressive. For example `hasApiTab`.

omikron-fp:
Broken styles look like this
tab-styling

damrath:
2018-09-03-141444_1050x464_scrot

Using `<paper-tab>`'s `disabled` property. If no api is avail. the tab will be grayed out and the tab is not selectable

PR #59, opened 2018-09-05
[#58] Export Integration

### LINKED ISSUE #58: Export Integration
#### Expected Behavior:

1. A script is added, that will run the [documentation-export](https://github.com/FACT-Finder-Web-Components/documentation-export) when executed as a jenkins build-step
2. The exported `.csv` files are served with the app.

damrath:
fixes #58


run-export.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+DOCS=markdown/en
+API=${DOCS}/api
+OF=$(pwd)/export
```
KrzysztofKozlowskiOBSS: Is absolute path required for export? If yes, can we use $PWD instead of evaluating pwd command?

run-export.sh
```diff
@@ -0,0 +1,15 @@
...
+
+if [ ! -d "$OF" ]; then
+    mkdir -p "$OF"
+fi
+
+if [ -d "$API" ]; then
+    java -jar ./documentation-export.jar -d "$DOCS" -a "$API" -o "$OF"
+else
+    exit 1;
```
KrzysztofKozlowskiOBSS: echo some error message here

Issue #77, opened 2018-09-20
Search result look & feel


damrath:
#### Current behaviour

The search results are currently displayed as cards (bootstrap) with 
cursor pointer on hover but don't really have the L&F of a clickable item.

#### Expected Behaviour

Adding a hover animation like in the [FF-Documentation Search](http://search.doku.fact-finder.de/ff-docu-search/index.html)

PR #78, opened 2018-09-20
[#77] hover box shadow animation and some formatting

damrath:
- box-shadow with ease in when result is hovered
- adjusted the colors to match the FF blue
- properties/mixins (labels) are now in monospaced font


damrath:
2018-09-20-114841_1908x1162_scrot

PR #86, opened 2018-09-26
[#72][FIX] use scrollIntoView() instead of anchors

### LINKED ISSUE #72: links to event types are unexpected
#### Current behaviour
All links within "List of all event types" on https://web-components.fact-finder.de/api/core-event-aggregator#tab=docs go to 404 or basically https://web-components.fact-finder.de/home

#### Expected Behaviour
Go to somewhere with some details to the clicked event

damrath:
fixes #72


damrath:
@torsten-knauf it would work, but would require changing the `#tab=tab-name` "url-parameter". you should open a new issue for that, requesting that feature.

Issue #68, opened 2018-09-10
keep-url-params should be better explained


jaroslawweselski:
#### Current behaviour

On https://web-components.fact-finder.de/api/ff-searchbox#tab=api you can find keep-url-params parameter, which has incomprehensible name and description. Client doesn't know what should be in this parameter.

#### Expected Behaviour

Not required but recommended may be change of parameter name. Description should be updated, recommended examples e.g.:

With this property you can define a whitelist for HTTP parameters not related to the Fact-Finder itself. Only these parameters were be allowed.
Example: parameter1,parameter2

(not sure if ',' is separator here, please check)

PR #92, opened 2018-09-27
[#68 #91] keep-url-params and tracking docs

### LINKED ISSUE #91: tracking-with-js `event` property
[tracking-with-js](https://web-components.fact-finder.de/documentation/tracking-with-js)
```Javascript
track.checkout({
    channel: "bergfreunde-co-uk",
    sid: "VqneWatI3sYNaCMmeNqRPGxcUTQag2",
    id: "010-0242-0318",
    price: 95.95,
    masterId: "9478a3507589b60ff7a6234a901298ec",
    count: 99,
    event: "cart"
});
```
`event` property is always overridden in all tracking functions

Expected:

property shouldn't be mentioned

damrath:
fixes #91 #68


markdown/en/api/ff-searchbox.api.md
```diff
@@ -41,12 +41,12 @@ ___
 | **add-params**&nbsp;(String) (default: empty)| _The parameter string has to be URL-encoded._ With this property you can deliver standard parameters which will then be attached to the search request. Example: add-params="param1=abcd,param2=xyz"|
 | **add-tracking-params**&nbsp;(String) (default: empty)| With this property you can deliver standard parameters which are attached to every tracking request. Example: add-tracking-params="param1=abcd,param2=xyz"|
 | **keep-filters**&nbsp;(Boolean) **Options**: &nbsp;true, &nbsp;false (default: false)| With this property you can determine, if filters which where set before the search (e.g. via ASN) should be kept or discarded.|
-| **keep-url-params**&nbsp;(String) (default: empty)| Determines if parameters which are written into the URL should be kept.|
+| **keep-url-params**&nbsp;(String) (default: empty)| Comma separated values; specifies the url-parameters that should be included in requests to FACT-Finder. If this is set to `all`, every parameter will be used |
```
omikron-fp: Would it be helpful to explicitly mention FACT-Finder unrelated paramters here? This is a major use case of this attribute, isn't it?

damrath: It pertains to ALL attributes, non-FF and FF alike. But we could mention it. I'm not sure. What do you think?

damrath: @toarm what do you think?

omikron-fp: I'm imagining users who have a setup that requires their own parameters. And they have a question in their mind like _How can I keep my own, FACT-Finder unrelated parameters?_
I personally appreciate buzzwords in documentation. Makes it easier for me to scan. Even `Ctrl+F` might yield some results.
In general, documentation doesn't have to be normalized. Repetition and explicitness are good. So it becomes a piece of empathetic customer care instead of _Amt_ where you only get answers for precisely what you asked.

_A common dialogue I have:_
Me: "I don't eat animals."
Other person: "Oh really? What about fish?"
Me: "..... _(aren't fish animals?)_"

So I'm always a bit careful with _ALL_. It's seems just too much or too implicit to some people.

damrath: Very well. I will add this use case to the docs.

*Leave it to a vegetarian to bring that up* :wink:

markdown/en/api/ff-searchbox.api.md
```diff
@@ -41,7 +41,7 @@ ___
...
 | **add-tracking-params**&nbsp;(String) (default: empty)| With this property you can deliver standard parameters which are attached to every tracking request. Example: add-tracking-params="param1=abcd,param2=xyz"|
 | **keep-filters**&nbsp;(Boolean) **Options**: &nbsp;true, &nbsp;false (default: false)| With this property you can determine, if filters which where set before the search (e.g. via ASN) should be kept or discarded.|
-| **keep-url-params**&nbsp;(String) (default: empty)| Comma separated values; specifies the url-parameters that should be included in requests to FACT-Finder. If this is set to `all`, every parameter will be used |
+| **keep-url-params**&nbsp;(String) (default: empty)| Comma separated values; specifies the url-parameters that should be included in requests to FACT-Finder. *This is particularly useful if you only want to keep your own custom non-FF attributes in the URL.* If this is set to `all`, every parameter will be used |
```
omikron-fp: Really? If you "**only** want to keep your own"? Sounds like it would throw away other parameters. Does it?

Would you also consider placing it at the end of the functional explanation? So it becomes like a side note.

markdown/en/api/ff-searchbox.api.md
```diff
@@ -41,7 +41,7 @@ ___
...
 | **add-tracking-params**&nbsp;(String) (default: empty)| With this property you can deliver standard parameters which are attached to every tracking request. Example: add-tracking-params="param1=abcd,param2=xyz"|
 | **keep-filters**&nbsp;(Boolean) **Options**: &nbsp;true, &nbsp;false (default: false)| With this property you can determine, if filters which where set before the search (e.g. via ASN) should be kept or discarded.|
-| **keep-url-params**&nbsp;(String) (default: empty)| Comma separated values; specifies the url-parameters that should be included in requests to FACT-Finder. *This is particularly useful if you only want to keep your own custom non-FF attributes in the URL.* If this is set to `all`, every parameter will be used |
+| **keep-url-params**&nbsp;(String) (default: empty)| Comma separated values; specifies the url-parameters that should be included in requests to FACT-Finder. If this is set to `all`, every parameter will be used. *This is particularly useful if you want to keep your own custom attributes in the URL whilst hiding the FACT-Finder url parameters with* `use-paramter="false"` |
```
omikron-fp: Sounds nice!
Looks like there is a type in `use-paramter`.
Where does that attribute actually come from? There is a `use-url-parameter` documented.
While at it, there is another typo "_serach_" in the document. Maybe you could fix that, too.

When we are talking about _parameters_ here, is `if you want to keep your own custom attributes in the URL` actually correct? Should it be `parameters` here as well?

PR #99, opened 2018-10-09
[FFWEB-805] add pages for ff-middleware and ff-multi-attribute-parsing

omikron-fp:
Adds a page for `ff-middleware` and `ff-multi-attribute-parsing`

Doesn't have a demo. I'm wondering whether it would make sense at all to have one in this case.

### DIFF
--- a/markdown/en/ff-middleware.md
+++ b/markdown/en/ff-middleware.md
@@ -1,0 +1,127 @@
+## Middleware
+
+Starting from version `1.2.12` FACT-Finder Web Components support the concept of middleware. You can register and configure pre-defined modules to manipulate the data being exchanged between Web Components and FACT-Finder.
+Middleware modules can be registered through custom elements or directly through JavaScript.
+
+## Custom Elements Approach
+
+To register middleware modules through custom elements place the `ff-middleware` element directly inside the `ff-communication` element. It is a container for all modules that shall be configured. 
+
+```html
+<ff-communication url="..."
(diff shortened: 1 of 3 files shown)


markdown/en/ff-middleware.md
```diff
@@ -0,0 +1,127 @@
+## Middleware
+
+Starting from version `1.2.12` FACT-Finder Web Components support the concept of middleware. You can register and configure pre-defined modules to manipulate the data being exchanged between Web Components and FACT-Finder.
```
damrath: I'd prefer `Starting with version ...` instead of `Starting from version ...`.

omikron-fp: Pretty difficult one. And quite difficult to find some nice information on this topic.
I liked _PaulQ's_ answer of [this thread](https://forum.wordreference.com/threads/start-from-vs-start-with.2546655/) quite much. In that case I would favour `Starting from` because to me a version number feels like a part of a sequence from with we are planning to move away.

This [StackExchange](https://english.stackexchange.com/questions/112972/starting-with-vs-starting-from) thread is maybe not so suitble for our context but might solidify the favour for one or the other nevertheless.

damrath: I see your point, it makes sense. The other version just sounds more natural to *me*.

omikron-fp:
Rebased and force pushed due to implementation of #100 (add noDemo flag)

PR #103, opened 2018-10-09
[#98] fix param type

### LINKED ISSUE #98: use-cache attribute of ff-communication has wrong type in docu
`ff-communication` description in https://web-components.fact-finder.de/api/ff-searchbox#tab=api states that `use-cache` is a String attribute, but in src-code it is a Boolean. 

Because Boolean and String attributes are treated fundamental different ( [s. here](https://web-components.fact-finder.de/documentation/attribute-basics) ) we should fix the docu.

damrath:
closes #98


damrath:
Almost every line had some mistake in it :weary:

PR #105, opened 2018-10-10
Ffweb 860 js configuration of ff communication

torsten-knauf:
I included `search-immediate` because it was already there. What do you think about removing it from the "3. Configuration" section and move it to "4. Your First Search". My reasoning is, that it is useful for first search and things like category pages, but it is to much information for minimal setup. Specially with the note in this section, that you don't always want to use it, it feels missplaced there to me.

If you agree, I would do that in another future ticket / branch.

### DIFF
--- a/markdown/en/documentation/configuration.md
+++ b/markdown/en/documentation/configuration.md
@@ -18,7 +18,7 @@ Specify the FACT-Finder version. E.g. `7.2` or  `7.3`.
 ---
 
 #### **channel** 
-Specify the channel name which is used for the search. You can check which channels are available at the FACT-Finder UI.
+Specify the name of the channel which is used for the search. You can find the available channels in the FACT-Finder UI.
 
 ---
 
@@ -45,11 +45,11 @@ We recommend placing the `ff-communication` element immediately after the `body`
 ```
 
 #### **search-immediate**
-The `search-immediate` attribute ensures, that a search is automatically triggered as soon as `ff-communication` was successfully attached to the DOM. The triggered search request uses the current URL parameters. If no parameters are present, it uses `*` as query (search for all products) in conjunction with the other settings. 
+The `search-immediate` attribute ensures, that a search is automatically triggered as soon as `ff-communication` was successfully attached to the DOM. The triggered search request uses the current URL parameters in conjunction with the settings provided by `ff-communication`-attributes. If no search query is provided through the url or an attribute `*` will be used as default (search for all products). 
 
 **NOTE**
 
-You don't want the `search-immediate` attribute to be present on your landing (home) page because this would trigger a search every time a user enters your shop. Usually you want to use it only on category pages and the search result page.
+You don't want the `search-immediate` attribute to be present on your landing (home) page because this would trigger a search every time a user enters your shop. Usually you only want to use it on category pages and the search result page.
 
 
 ## JavaScript Approach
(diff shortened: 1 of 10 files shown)


markdown/en/documentation/configuration.md
```diff
@@ -1,8 +1,28 @@
 ## Configuration
 
 ---
-This chapter describes the minimum configuration necessary to search and retrieve results from FACT-Finder.
+This chapter describes the minimal configuration necessary to search and retrieve results from FACT-Finder. You can configure the communication via the custom element `ff-communication` or directly via JavaScript.
 
+### A minimal configuration consists of the following settings:
+
+#### **url**
+Specify the FACT-Finder endpoint. Please note the context name `/FACT-Finder-7.2`. It is not sufficient to provide the top-level domain like `http://web-components.fact-finder.de/` which will result in an error.
```
damrath: use generic URL like `https://myshop.fact-finder.de/`

markdown/en/documentation/configuration.md
```diff
@@ -1,8 +1,28 @@
...
+
+**Note:** Make sure you are using the same protocol (http://, https://) for the `url` setting and webapp. Using `url="http://web-components.fact-finder.de/FACT-Finder-7.2"` on a page hosted via **HTTPS** will result in an error.
```
damrath: use the generic URL

markdown/en/documentation/configuration.md
```diff
@@ -1,8 +1,28 @@
...
+
+---
+
+#### **version**
+Specify the FACT-Finder version. E.g. `7.2` or  `7.3`.
+
+---
+
+#### **channel** 
+Specify the channel name which is used for the search. You can check which channels are available at the FACT-Finder UI.
```
damrath: `Specify the name of the channel you want to use for the search. You can find the available channels in the FACT-Finder UI.`

markdown/en/documentation/configuration.md
```diff
@@ -18,39 +38,38 @@ This chapter describes the minimum configuration necessary to search and retriev
...
-#### **url**
-Specify the FACT-Finder endpoint. Please note the context name `/FACT-Finder-7.2`. It is not sufficient to provide the top-level domain like `http://web-components.fact-finder.de/` which will result in an error.
-
-**Note:** Make sure you are using the same protocol (http://, https://) for the `url` attribute and webapp. 
-
-Using `<ff-communication url="http://web-components.fact-finder.de/FACT-Finder-7.2" ...>` on a page hosted via **HTTPS** will result in an error.
-
----
+#### **search-immediate**
+The `search-immediate` attribute ensures, that a search is automatically triggered as soon as `ff-communication` was successfully attached to the dom. The triggered search request uses the current URL parameters. If no parameters are present, it uses `*` as query (search for all products) in conjunction with the other settings.
```
damrath: dom -> DOM
The triggered search request uses `ff-communication` parameters. If a parameter isn't set, default settings are used. The default search term is `*` (search for all products).

markdown/en/documentation/configuration.md
```diff
@@ -18,39 +38,38 @@ This chapter describes the minimum configuration necessary to search and retriev
...
----
+#### **search-immediate**
+The `search-immediate` attribute ensures, that a search is automatically triggered as soon as `ff-communication` was successfully attached to the dom. The triggered search request uses the current URL parameters. If no parameters are present, it uses `*` as query (search for all products) in conjunction with the other settings. 
 
-#### **version**
-Specify the FACT-Finder version. E.g. 7.2, 7.3 and so forth.
+**NOTE**
 
----
+You don't want the `search-immediate` attribute to be present on your landing (home) page because this would trigger a search every time a user enters your shop. Usually you want to use it only on category pages and the search result page.
```
damrath: Usually you *only* want to use it on category pages and the search result page.

markdown/en/documentation/configuration.md
```diff
@@ -18,39 +38,38 @@ This chapter describes the minimum configuration necessary to search and retriev
...
+    /* after this minimal configuration, a search can be triggered manual e.g. through */
+    factfinder.communication.FFCommunicationEventAggregator.addFFEvent({
+         type: "search",
+         query: "*"
+    });
+});
+``` 
 
 **NOTE**
+To prevent our custom elements from missing a factfinder response, before they where loaded, dispatching of results to subscribers is cached and deferred until all custom elements have loaded. When all custom elements have loaded, `ff-communication` calls `factfinder.communication.ResultDispatcher.startDispatching()`. If `ff-communication` is not used, `factfinder.communication.ResultDispatcher.startDispatching()` has to be called manually in JavaScript.
```
damrath: replace "custom elements" with "Web Components" so it's more consistent with the rest of the documentation
... before they're ...
... until all Web Components are ready to receive the response. As soon as Web Components have finished loading `ff-communication` calls ...

torsten-knauf: In this case, we should rename `custom elements` through `Web Components` in [#99](https://github.com/FACT-Finder-Web-Components/documentation/pull/99) as well. @omikron-fp  What do you think?

omikron-fp: We could do that.
I'm wondering though if there is an understanding out there that maps `Web Components` to the whole thing as a framework. In which case it might be more obvious to distinguish between everything as a whole called `Web Components` and the actual HTML elements called `custom elements`.
Is this a real thing or is this just me?

torsten-knauf: I think that the technology we use is `Web Components` and our library provides `custom elements` everyone can use. Hence I would tend to change most cases to `custom elements` :P 

I will create an own issue for more detailed consideration of that topic and skip it in this pull request

PR #129, opened 2018-11-21
Multi-version support

KrzysztofKozlowskiOBSS:
Polymer serve is not handling dots in URLs properly so we have to `polymer build` and run app via Express until we find a workaround.


run-export.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-DOCS=markdown/en
+DOCS=markdown/3.0/en
```
torsten-knauf: why do we not have to export 1.2? (just aksing, I have no idea what this file does)

KrzysztofKozlowskiOBSS: It's a temporary thing, we have to implement multi-version support in our docs export tool and on the FACT-Finder side.

PR #131, opened 2018-11-27
add mixin section to ff-header-navigation for nav-elements' anchors

omikron-fp:
Documentation of the latest bugfix in `ff-header-navigation`.
If `ff-header-navigation` gets migrated to LitElement before the `3.0` release, this change is not relevant.


torsten-knauf:
What is your opinion on mergen now or wait for release? I think merging now and remove it, after someone has migrated navigation is fine.

PR #141, opened 2018-12-04
[#140] use static serve

### LINKED ISSUE #140: Serve multi-version export files
New exports need download URLs

damrath:
fixes #140 
serve export files using express' static serve

### DIFF
--- a/app.js
+++ b/app.js
@@ -2,25 +2,15 @@
 
 const express = require('express');
 const app = express();
-const exportPath = '/export/';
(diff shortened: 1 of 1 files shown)


torsten-knauf:
The change looks good, but I am afraid I don't know how to test it. Where is the download of those _.csv_-files purposed to happen?

torsten-knauf:
The direct url downloading works as expected. Is there any other thing to check?

PR #153, opened 2019-01-25
FFWEB-958 upgrade ff-searchbox and ff-searchbutton documentation

wdutkiewicz-obj:
Upteded documentation of ```ff-searchbox``` and ```ff-searchbutton``` to version 3.0 of these components in all applicable areas.

### DIFF
--- a/markdown/3.0/en/api/ff-searchbox.api.md
+++ b/markdown/3.0/en/api/ff-searchbox.api.md
@@ -8,7 +8,7 @@ ___
 | **select-onclick**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "false") | Determines if the contents of the search box should be selected if the inside of the box is clicked. |
 | **use-suggest**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "true") | Determines if a request should be sent to the suggest interface as soon as a user starts typing in a search term. |
 | **suggest-delay** &nbsp;(Number) (default: 0) | Triggers a suggest request only after the delay expired and no more input is set. When the input field changes in the delay time frame the delay will be reset. |
-
+| **value**&nbsp;(String) (default: empty) | Gets or sets the value of related HTML `<input>` element. This property cannot be set on `<ff-searchbox>` element and should only be used in custom JavaScript code. Setting value triggers `input` event on `<input>` element. Therefore suggestions are shown (if enabled). |
 ### Events
 | Name | Description |
 | ---- | ----------- |
@@ -17,11 +17,15 @@ ___
 ### Methods
 | Name | Description |
 | ---- | ----------- |
-| **search** | Initializes a search.|
+| **search()** | Initializes a search.|
+| **resetInput(selector = 'input')** | Sets the `input` reference to the nested HTML element pointed by selector. |
 
 ## `ff-searchbutton`
 ___
-
+### Methods
+| Name | Description |
+| ---- | ----------- |
+| **resetButton(selector = 'button')** | Sets the `button` reference to the nested HTML element pointed by selector. |
 
 ## `ff-communication`
 ___
--- a/markdown/3.0/en/api/ff-searchbox.api.md
+++ b/markdown/3.0/en/api/ff-searchbox.api.md
@@ -8,7 +8,7 @@ ___
 | **select-onclick**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "false") | Determines if the contents of the search box should be selected if the inside of the box is clicked. |
 | **use-suggest**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "true") | Determines if a request should be sent to the suggest interface as soon as a user starts typing in a search term. |
 | **suggest-delay** &nbsp;(Number) (default: 0) | Triggers a suggest request only after the delay expired and no more input is set. When the input field changes in the delay time frame the delay will be reset. |
-| **value**&nbsp;(String) (default: empty) | Gets or sets the value of related HTML `<input>` element. This property cannot be set on `<ff-searchbox>` element and should only be used in custom JavaScript code. Setting value triggers `input` event on `<input>` element. Therefore suggestions are shown (if enabled). |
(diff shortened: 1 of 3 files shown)


markdown/3.0/en/api/ff-searchbox.api.md
```diff
@@ -8,7 +8,7 @@ ___
 | **select-onclick**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "false") | Determines if the contents of the search box should be selected if the inside of the box is clicked. |
 | **use-suggest**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "true") | Determines if a request should be sent to the suggest interface as soon as a user starts typing in a search term. |
 | **suggest-delay** &nbsp;(Number) (default: 0) | Triggers a suggest request only after the delay expired and no more input is set. When the input field changes in the delay time frame the delay will be reset. |
-
+| **value**&nbsp;(String) (default: empty) | Gets or sets the value of related HTML `<input>` element. This property cannot be set on `<ff-searchbox>` element and should only be used in custom JavaScript code. Setting value triggers `input` event on `<input>` element. Therefore suggestions are shown (if enabled). |
```
torsten-knauf: I had to read this 3 times to understand why it cannot be set and should be set at the same time.

Maybe something like:

This property cannot be set as attribute on `<ff-searchbox>` element. Though JavaScript the value of the  related `<input>` element can be get and set. Setting the value triggers an `input` event on the `<input>` element. Therefore suggestions are shown if enabled.

markdown/3.0/en/ff-searchbox.md
```diff
@@ -24,58 +24,73 @@ The following code-example shows the aforementioned configuration.
...
-<input is="ff-searchbox" />
+<ff-searchbox>
+    <input />
+</ff-searchbox>
 ```
 
 ### Placeholder
-Because the `ff-searchbox` is an extension to a normal `<input>` you can
-use all the standard attributes of `<input>` such as `placeholder`.
+Because the `ff-searchbox` does not limit features o HTML it wraps,
```
torsten-knauf: o**f** HTML

But maybe just remove the first part, because why shouldn't it be that way?

On second part, maybe even remove the full paragraph? It is standard HTML, isn't it?

markdown/3.0/en/ff-searchbox.md
```diff
@@ -24,58 +24,73 @@ The following code-example shows the aforementioned configuration.
 ```
 For more information, see the [API reference](/api/ff-searchbox#tab=api).
 
-## Extending the input
-The `ff-searchbox` extends a normal `<input>`. Simply add the attribute
-`is="ff-searchbox"` to a standard `<input>` element. On ENTER a search
-query is fired with the current value.
+## Wrapping input element with ff-searchbox
+The `ff-searchbox` wraps a regular HTML `<input>` element. Simply place
+`<input>` between `<ff-searchbox>` tags. On ENTER a search query is fired
```
torsten-knauf: maybe add "...a search query is fired _to FACT-Finder_"

markdown/3.0/en/ff-searchbox.md
```diff
@@ -24,58 +24,73 @@ The following code-example shows the aforementioned configuration.
...
+<ff-searchbox select-onclick="true">
+    <input />
+</ff-searchbox>
+
 ```
 
 ## Adding Suggest
 If you want a suggest functionality on your page you can set the
-`use-suggest` attribute to **true** *(default is true)*. You also need
+`use-suggest` attribute to `true` *(default is `true`)*. You also need
```
torsten-knauf: the type used to be String and not Boolean. With the new StringyBoolean I don't know what we should write in the documentation. But it should be consistent with the type in the Api tab (there it is String) and the example below (there it is also String). So probably `"true"` instead of `true` :)

markdown/3.0/en/ff-searchbox.md
```diff
@@ -24,58 +24,73 @@ The following code-example shows the aforementioned configuration.
...
-<input is="ff-searchbox" suggest-onfocus="true" />
+<ff-searchbox suggest-onfocus="true">
+    <input />
+</ff-searchbox>
 ```
 
 ### Hiding Suggest
 Usually, you want the suggest to disappear once the user clicks
-somewhere else. With the attribute `hidesuggest-onblur` set to **false**
+somewhere else. With the attribute `hidesuggest-onblur` set to `false`
```
torsten-knauf: I think it is a String; hence `"false"`

markdown/3.0/en/ff-searchbox.md
wdutkiewicz-obj: We both know it's standard HTML, but end user does not necessarily does. For some reason this section was there and I'd rather keep it.

wdutkiewicz-obj:
> See comments:
> 
> In http://127.0.0.1:8081/api/3.0/ff-suggest#tab=docs is also an overseen `<input is="ff-searchbox"`

I fixed it in ff-suggest document update branch.

markdown/3.0/en/api/ff-searchbox.api.md
wdutkiewicz-obj: I edited the description. Please check now.

markdown/3.0/en/ff-searchbox.md
wdutkiewicz-obj: "... a search query to FACT-Finder is fired..." - please check now.

markdown/3.0/en/ff-searchbox.md
wdutkiewicz-obj: I wrapped string values with double quotes - please check now.

markdown/3.0/en/ff-searchbox.md
wdutkiewicz-obj: Wrapped with double quotes.

markdown/3.0/en/ff-searchbox.md
torsten-knauf: > o**f** HTML

sorry, seems that this was a to lazy hint from me. Leaving the sentence there is fine with me, but please spend a **f** to  the lonely _o_ ;)

PR #154, opened 2019-01-25
FFWEB-968 upgrade ff-suggest documentation

wdutkiewicz-obj:
Updated documentation of ```ff-suggest``` to version 3.0.

### DIFF
--- a/markdown/3.0/en/ff-suggest.md
+++ b/markdown/3.0/en/ff-suggest.md
@@ -14,10 +14,9 @@ The `ff-suggest` is triggered when at least 2 characters are in the
 search box.
 
 ```html
-<input is="ff-searchbox"
-       placeholder="Search..."
-       suggest-onfocus="false"
-       hidesuggest-onblur="true"/>
+<ff-searchbox suggest-onfocus="false" hidesuggest-onblur="true">
+    <input placeholder="Search..." />
+</ff-searchbox>
 ```
 
 For more information on `ff-searchbox`, see
@@ -53,7 +52,7 @@ at this position and will be an exact copy of this element. `ff-suggest-item` su
 
 Repeat this process for all configured `"suggestTypes"`.
 
-```html 
+```html
 <ff-suggest>
     <section>
         <div data-container="suggestType">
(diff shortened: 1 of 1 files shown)


markdown/3.0/en/ff-suggest.md
```diff
@@ -81,6 +80,49 @@ rather than the rendered: "**Back**packs".
...
+### Rendered HTML
+
+Container templates provided inside `<ff-suggest>` tag will be rendered inside
+several additional wrapping `<div>` HTML elements.
+
+For example, the following HTML provided by the user:
+
+```html
+<ff-suggest layout="block">
+    <section>
```
torsten-knauf: I think a not empty section would be more informative.

Maybe refer to the `ff-suggest` container from above's section and show an example of its rendered output?

markdown/3.0/en/ff-suggest.md
```diff
@@ -127,9 +169,9 @@ event `"suggest-item-clicked"`.
 
 See the following example:
 
-```html 
+```html
 <ff-suggest>
-    //content...
+    <!--content-->
```
torsten-knauf: 👍

markdown/3.0/en/ff-suggest.md
wdutkiewicz-obj: I updated HTML with a section with some sample content.

markdown/3.0/en/ff-suggest.md
```diff
@@ -81,16 +80,64 @@ rather than the rendered: "**Back**packs".
...
+```html
+<ff-suggest layout="block">
+    <div style="position:relative;width:100%">
+        <div class="ffw-suggestContainerWrapper">
+            <div class="ffw-suggestContainer ffw-blockLayout">
+                <section>
+                    <div data-container="suggestType">
+                        <span>Suggestions for suggestType</span>
+                        <div>
+                            <ff-suggest-item type="suggestType">
```
torsten-knauf: Mh, from comment in src code I would guess, that _"suggestType"_ is only a placeholder. I think we should list the possible types somewhere and create a real example without placeholder. Because it wasn't done before, feel free to decide, if you want to do it in this PR or if I should create a new ticket for that
```
/**
             * The "<b>type</b>" attribute specifies the type of the element. Valid type values are:<br>
             * <ul>
             *     <li><b>searchTerm</b><br></li>
             *     <li><b>brand</b><br></li>
             *     <li><b>productName</b><br></li>
             *     <li><b>category</b><br></li>
             *     <li><b>all</b></li>
             * </ul>
```

wdutkiewicz-obj: As far as I know the set of suggest types is not closed. These are predefined values - user can add nev suggest type in FACT-Finder and map it to DB field. The types you found in source code comment are listed in ff-suggest-item documentation. If you want this section of documentation to be updated in details, please create a new ticket as this requires some more time.

PR #158, opened 2019-01-29
Include paging dropdown in paging

torsten-knauf:
This little change makes the sidebar more overseenable and the docu more structured

Related demo update is in this PR https://github.com/FACT-Finder-Web-Components/demos/pull/9


torsten-knauf:
Note at myself: Don't forget to merge https://github.com/FACT-Finder-Web-Components/demos/pull/9 afterwards

KrzysztofKozlowskiOBSS:
To be honest, I prefer them separated. Regular paging is a very popular component, it's like a must have for every shop. On the other hand, dropdown is rarely used. I think we shouldn't bother users with dropdown, when they just want to integrate regular paging.
If you want to make sidebar shorter, there are many super small campaign pages that could be merged.

torsten-knauf:
> To be honest, I prefer them separated. Regular paging is a very popular component, it's like a must have for every shop. On the other hand, dropdown is rarely used. I think we shouldn't bother users with dropdown, when they just want to integrate regular paging.

The dropdown is at the end, so if they are already happy they don't have to bother with it. My point of having one section for _paging_ is, that it increases the structure of the sidebar, to have one entry for one kind of feature / element.

@toarm @omikron-fp @wdutkiewicz-obj @damrath Any opinion for outvoting someone? 🙃 

> If you want to make sidebar shorter, there are many super small campaign pages that could be merged.

👍 Yes! Something like that is on my todo for campaign-migration. I have something like a foldable menu _camapaigns_ with the specific campaigns as entries in my mind. We will see how this will work out :) - But my main motivation is not to shorten the sidebar but to increase the clarity of it (ok, that is somehow close related to shortening^^)

torsten-knauf:
seems to be a point for a more general overwork of common abstraction layer and UI experience. Hence closing for now and putting it on my long term todo

PR #159, opened 2019-02-06
Ffweb 962 upgrade ff navigation

wdutkiewicz-obj:
Updated documentation for ff-navigation element.

### DIFF
--- a/markdown/3.0/en/ff-navigation.md
+++ b/markdown/3.0/en/ff-navigation.md
@@ -5,14 +5,14 @@ To get a simple `ff-navigation` started, you only need to use the following code
 ```html
 <ff-navigation >
     <ff-navigation-item>
-        <div slot="name">{{name}}</div>
+        <div data-content="item">{{name}}</div>
     </ff-navigation-item>
 </ff-navigation>
 ```
 
-You need to have at least one `ff-navigation-item` defined. The `ff-navigation-item` is used as a template for all the elements of the Navigation. We use the term 'elements' to describe a single navigation category in a tree-like structure.
+You need to define least one `ff-navigation-item`. The `ff-navigation-item` is used as a template for all the elements of the Navigation. We use the term 'elements' to describe a single navigation category in a tree-like structure.
 
-In this `ff-navigation-item` you have access to the element's values. Such as `'name'` and `'recordCount'`. We define a `div` for the slot 'content' in which you define how each element of the navigation should be displayed. You can add a css class on this element:
+In scope of `ff-navigation-item` you have access to the navigation element's values, such as `name` and `recordCount`. We define a `div` for the item template in which you define how each element of the navigation should be displayed. You can add a css class on this element:
 
 ```html
 <style>
@@ -24,18 +24,20 @@ In this `ff-navigation-item` you have access to the element's values. Such as `'
 </style>
 <ff-navigation>
     <ff-navigation-item>
-        <div slot="name" class="my-item">{{name}}</div>
+        <div data-content="item" class="my-item">{{name}}</div>
     </ff-navigation-item>
 </ff-navigation>
 ```
 
+**NOTICE:** `data-content` attribute is mandatory and its value should be set to `"item"`. If omitted, menu will not render correctly.
+
 ## Templates
 
-You can also define 3 possible templates, which are used in different situations. To define these set the `'type'` property on the `ff-navigation-item`. A `ff-navigation-item` with no type is used as the **default** fallback template and has to be defined.
+You can also define 3 possible templates, which are used in different situations. To define these set the `type` property on the `ff-navigation-item`. A `ff-navigation-item` with no type is used as the **default** fallback template and has to be defined.
 
 ### 'header'
 
-The **header** template is used in your first level of the navigation. This allows you to style the Header differently than all following elements. You can define different Layouts inside the navigation element. For example, if you want to display the `'recordCounts'` inside a normal element, but not in the header:
+The **header** template is used in your first level of the navigation. This allows you to style the Header differently than all following elements. You can define different Layouts inside the navigation element. For example, if you want to display the `'recordCount'` inside a normal element, but not in the header:
 
 ```html
 <style>
@@ -53,11 +55,11 @@ The **header** template is used in your first level of the navigation. This allo
 </style>
 <ff-navigation>
     <ff-navigation-item>
-        <div slot="name" class="my-item">{{name}} ({{recordCount}})</div>
+        <div data-content="item" class="my-item">{{name}} ({{recordCount}})</div>
     </ff-navigation-item>
 
     <ff-navigation-item type="header">
-        <div slot="name" class="my-header">{{name}}</div>
+        <div data-content="item" class="my-header">{{name}}</div>
     </ff-navigation-item>
 </ff-navigation>
 ```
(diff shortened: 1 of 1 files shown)


torsten-knauf:
I was so free to change the PR target from _master_ to _release/3.0_

markdown/3.0/en/ff-navigation.md
```diff
@@ -5,14 +5,14 @@ To get a simple `ff-navigation` started, you only need to use the following code
...
 <ff-navigation >
     <ff-navigation-item>
-        <div slot="name">{{name}}</div>
+        <div data-content="item">{{name}}</div>
     </ff-navigation-item>
 </ff-navigation>
 ```
 
-You need to have at least one `ff-navigation-item` defined. The `ff-navigation-item` is used as a template for all the elements of the Navigation. We use the term 'elements' to describe a single navigation category in a tree-like structure.
+You need to define least one `ff-navigation-item`. The `ff-navigation-item` is used as a template for all the elements of the Navigation. We use the term 'elements' to describe a single navigation category in a tree-like structure.
```
torsten-knauf:
```suggestion
You need to define at least one `ff-navigation-item`. The `ff-navigation-item` is used as a template for all the elements of the Navigation. We use the term 'elements' to describe a single navigation category in a tree-like structure.
```

markdown/3.0/en/ff-navigation.md
```diff
@@ -5,14 +5,14 @@ To get a simple `ff-navigation` started, you only need to use the following code
...
 
-In this `ff-navigation-item` you have access to the element's values. Such as `'name'` and `'recordCount'`. We define a `div` for the slot 'content' in which you define how each element of the navigation should be displayed. You can add a css class on this element:
+In scope of `ff-navigation-item` you have access to the navigation element's values, such as `name` and `recordCount`. We define a `div` for the item template in which you define how each element of the navigation should be displayed. You can add a css class on this element:
```
torsten-knauf:
```suggestion
Within `ff-navigation-item` you have access to the navigation element's values, such as `name` and `recordCount`. We define a `div` for the item template in which you define how each element of the navigation should be displayed. You can add a css class on this element:
```

markdown/3.0/en/ff-navigation.md
```diff
@@ -24,18 +24,20 @@ In this `ff-navigation-item` you have access to the element's values. Such as `'
...
+
 ## Templates
 
-You can also define 3 possible templates, which are used in different situations. To define these set the `'type'` property on the `ff-navigation-item`. A `ff-navigation-item` with no type is used as the **default** fallback template and has to be defined.
+You can also define 3 possible templates, which are used in different situations. To define these set the `type` property on the `ff-navigation-item`. A `ff-navigation-item` with no type is used as the **default** fallback template and has to be defined.
 
 ### 'header'
 
-The **header** template is used in your first level of the navigation. This allows you to style the Header differently than all following elements. You can define different Layouts inside the navigation element. For example, if you want to display the `'recordCounts'` inside a normal element, but not in the header:
+The **header** template is used in your first level of the navigation. This allows you to style the Header differently than all following elements. You can define different Layouts inside the navigation element. For example, if you want to display the `'recordCount'` inside a normal element, but not in the header:
```
torsten-knauf: nice catch 👍

torsten-knauf:
```suggestion
The **header** template is used in your first level of the navigation. This allows you to style the Header differently than all following elements. You can define different Layouts inside the navigation element. For example, if you want to display the `recordCount` inside a normal element, but not in the header:
```

markdown/3.0/en/ff-navigation.md
```diff
@@ -193,16 +195,16 @@ In the **horizontal** layout, the 'level 0' or 'header' elements extend down. Al
...
 </ff-navigation>
 ```
 
 ## Responsiveness
 
-To make the `ff-navigation` responsive, you need to add more slots.
+To make the `ff-navigation` responsive, you need to add one more element.
 
-First you need to add a **menu** element to the `ff-navigation`. Only this element will be shown when the attribute 'mobile' of the `ff-navigation` is set to true. This element will internally get a click handler to display the first level of the navigation.
+First you need to add a **menu** (`data-content="menu"` attribute is required) element to the `ff-navigation`. Only this element will be shown when the attribute 'mobile' of the `ff-navigation` is set to true. This element will internally get a click handler to display the first level of the navigation.
```
torsten-knauf:
```suggestion
First you need to add a **menu** element with `data-content="menu"` attribute to the `ff-navigation`. Only this element will be shown when the attribute 'mobile' of the `ff-navigation` is set to `true`. This element will internally get a click handler to display the first level of the navigation.
```

markdown/3.0/en/ff-navigation.md
wdutkiewicz-obj: Within is related to time and distance at first. It means 'inside' as well. However, opening the sentence it doesn't look well. Scope is often used in IT, so in my opinion it fits in better our case.

wdutkiewicz-obj:
> 
> 
> I was so free to change the PR target from _master_ to _release/3.0_

Indeed, it was my mistake. I am sorry for that.

wdutkiewicz-obj:
> 
> 
> May I ask for your reasoning of changing `slot="name"` to `data-content="item" `?
> 
> I think this is an unnecessary breaking change. Besides I think `slot` is more appropriate - that is the placeholder where we insert the templates, isn't it?

The ff-navigation component has been modified so that it does not use the concept of shadow DOM and slots. I still need some way to mark content and container. Using ```data-container``` (in JavaScript code) and ```data-content``` (in HTML code) is more descriptive. ```slot="name"``` was be a remnant so I changed it. Besides, the documentation reflects current state of the component and it has already been approved and merged into release 3.0.

markdown/3.0/en/ff-navigation.md
wdutkiewicz-obj: Thank you.

markdown/3.0/en/ff-navigation.md
wdutkiewicz-obj: Fixed

markdown/3.0/en/ff-navigation.md
wdutkiewicz-obj: Fixed.

markdown/3.0/en/ff-navigation.md
wdutkiewicz-obj: Fixed.

torsten-knauf:
> The ff-navigation component has been modified so that it does not use the concept of shadow DOM and slots. I still need some way to mark content and container. Using `data-container` (in JavaScript code) and `data-content` (in HTML code) is more descriptive. `slot="name"` was be a remnant so I changed it. Besides, the documentation reflects current state of the component and it has already been approved and merged into release 3.0.

as far as I know, `data-content` is used to provide data - not to mark a node as "we will place it for you at the appropriate position in the dom" - isn't it? So I still think the previous `slot` is better naming.

Do I overlook something or could you have technically just used `const menu = this.querySelector('[slot="menu"]');` instead of this.querySelector('[data-content="menu"]'), without any breaking change for the customers template?
 
> Please, notice that removing shadow DOM and slots is just a middle step in the migration to Lit. We will probably change it so that additional attributes will not be necessary.

A reason more to not introduce this breaking change. As a customer I would not be to happy to have to change the same template twice within a short period.

But I think we don't have a standardized way to provide this insertion points yet and your source PR is already merged, so I would be OK with letting it that way. But I would appreciate it if you were motivated and rework this breaking change.

PR #164, opened 2019-02-20
FFWEB-941 upgrade ff-asn element

omikron-fp:
Updates `ff-asn` main and api pages.

### DIFF
--- a/markdown/3.0/en/api/ff-asn.api.md
+++ b/markdown/3.0/en/api/ff-asn.api.md
@@ -18,19 +18,12 @@ ___
 | ---- | ----------- |
 | **opened**&nbsp;(Boolean) | Determines if the filter group is expanded. If you want to have a group opened by default: `<ff-asn-group opened>` does the trick. |
 | **collapsible**&nbsp;(String)&nbsp;**Options**:&nbsp;"true",&nbsp;"false"&nbsp;(default: "true") | Determines, if the filter group is collapsible. |
-| **for-group**&nbsp;(String)&nbsp;(default: "true")| Determines, which filter group the template should be applied to. If set to `"all"`, the template is applied to all groups if no other template matches the requirements. |
+| **for-group**&nbsp;(String)&nbsp;(default: "all")| Determines, which filter group the template should be applied to. If set to `"all"`, the template is applied to all groups if no other template matches the requirements. |
 | **group**&nbsp;(Object) | The data for the filter group. |
 | **filter-style**&nbsp;(String) | With the filter-style property it is possible to use the ff-asn-group element as a template for all groups which match the filter style. (TREE, DEFAULT, MULTISELECT, SINGLESELECT) |
 | **lazy-load**&nbsp;(String) **Options**:&nbsp;"true",&nbsp;"false" (default: "true") | The lazy-load property defines if the ff-asn-elements for the hidden links container should be rendered when the asn group is dispatched or lazily just when the hiddenLinks container is opened via toggleHiddenLinksContainer() or showHiddenLinksContainer(). This improves performance for the first print and is by default set to true. |
 | **select-box**&nbsp;(String) **Options**:&nbsp;"true",&nbsp;"false" (default: "false") | Use this when the hiddenLinks should be an HTML select element |
 | **disable-auto-expand**&nbsp;(Boolean) | Prevents group from being automatically expanded when it contains any active filters. |
-
-### Mixins
-| Name | Description |
-| ---- | ----------- |
-| **--asn-all-links-container** | Is applied to the container of the ASN group. |
-| **--ff-asn-group-container** | Is applied to the container of the ASN group.|
-| **--asn-group-caption** | Is applied to the header of the ASN group. |
 
 ### Methods
 | Name | Description |
@@ -50,14 +43,14 @@ ___
 ### Properties
 | Name | Description |
 | ---- | ----------- |
-| **selected**&nbsp;(String)&nbsp;**Options**:&nbsp;"true",&nbsp;"false" (default: empty) | The alignment of the After Search Navigation. |
+| **selected**&nbsp;(String)&nbsp;**Options**:&nbsp;"true",&nbsp;"false" (default: "false") | The alignment of the After Search Navigation. |
 | **element**&nbsp;(Object)&nbsp;(default: empty) | The filter data. |
 | **group**&nbsp;(Object)&nbsp;(default: empty) | The filter group data. |
 
 ### Methods
 | Name | Description |
 | ---- | ----------- |
-| **clone** | Return a deep copy of the contained elements including all properties, behaviors, private fields, states and HTML templates. |
+| **clone(group, element)** | Returns a copy of the group element with the specified group and element to provide the new context. |
 
 ## `ff-asn-group-slider`
 ___
@@ -70,12 +63,6 @@ ___
 | **group**&nbsp;(Object) | The data for the filter group. |
 | **filter-style**&nbsp;(String) | Read only property, always set to "SLIDER" |
 | **disable-auto-expand**&nbsp;(Boolean) | Prevents group from being automatically expanded when it contains any active filters. |
-
-### Mixins
-| Name | Description |
-| ---- | ----------- |
-| **--all-links-container** | Is applied to the container of the ASN group. |
-| **--ff-asn-group-container** | Is applied to the container of the ASN group.|
 
 ### Methods
 | Name | Description |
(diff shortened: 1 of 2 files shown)


markdown/3.0/en/api/ff-asn.api.md
```diff
@@ -50,14 +43,14 @@ ___
...
-| **selected**&nbsp;(String)&nbsp;**Options**:&nbsp;"true",&nbsp;"false" (default: empty) | The alignment of the After Search Navigation. |
+| **selected**&nbsp;(String)&nbsp;**Options**:&nbsp;"true",&nbsp;"false" (default: "false") | The alignment of the After Search Navigation. |
 | **element**&nbsp;(Object)&nbsp;(default: empty) | The filter data. |
 | **group**&nbsp;(Object)&nbsp;(default: empty) | The filter group data. |
 
 ### Methods
 | Name | Description |
 | ---- | ----------- |
-| **clone** | Return a deep copy of the contained elements including all properties, behaviors, private fields, states and HTML templates. |
+| **clone(group, element)** | Returns a copy of the group element with the specified group and element to provide the new context. |
```
torsten-knauf: I wonder about the reason for making `clone` part of the public api - what is the use case, where a customer would use it?

I think in the long term the `clone` will be obsolete due to lit migration and generating HTML declarative  from data instead of this imperative `clone`. So I think _3.0_ would be a nice timing to drop that from public interface

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
 ## Using the After Search Navigation
-You can use the ASN by adding just one line to your HTML code.
+You can use the ASN by adding one line to your HTML code.
 ```html
 <ff-asn></ff-asn>
 ```
-If you do so, the ASN will look pretty much like nothing. So let's style it.
+Doing so will cause the ASN to use default HTML templates for all filters. You can customize the ASN's appearance by providing your own templates for its elements.
```
torsten-knauf: Mh, removing the _example with no effect_ would make it more comprehensive without losing any information, wouldn't it? 

Maybe change the text as followed:

```suggestion
Doing so will cause the ASN to use default HTML templates for all filters. You can customize the ASN's appearance by providing your own templates for its elements. Your customized templates must be wrapped within a `ff-asn-group` element. Below are covered some examples.
```

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
...
 
 ## Adding a filter group
 ```html
 <ff-asn>
     <ff-asn-group></ff-asn-group>
 </ff-asn>
 ```
-Adding a bare `ff-asn-group` does not change the styles but it lets us
-control how different kinds of `ff-asn-groups` should look.
+To customize filter groups you add `ff-asn-group` elements. However, adding an empty `ff-asn-group` as in the example above has no effect. You will likely want to specify your own HTML template.
```
torsten-knauf: see comment above

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
...
+
+
+## Changing the group caption
+If you want to change the group caption, you can annotate an HTML element with the `[slot="groupCaption"]` attribute. By doing so you can change the style and appearance. You can access the group's name through `{{group.name}}`. Here is an example of how a custom group caption could be implemented:
```
torsten-knauf:
```suggestion
If you want to change the group caption, you can annotate an HTML element with the `[slot="groupCaption"]` attribute. You can access the group's name through `{{group.name}}`. Here is an example of how a custom group caption could be implemented:
```

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
...
-acts as the default template for all groups that match this `[filter-style]`.
+```html
+<ff-asn-group>
+    <div data-container="detailedLinks">
+        <div data-content="detailedLinks">
+            <!-- always visible items will be added here -->
+        </div>
+    </div>
+    <div data-container="showMore">
+        <!-- The "showMore" container will only be rendered if there are "hiddenLinks" in the FACT-finder response.
```
torsten-knauf:
```suggestion
        <!-- The "showMore" container will only be rendered if there are "hiddenLinks" in the FACT-Finder response.
```

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
...
+        </div>
+    </div>
+    <div data-container="showLess">
+        <span class="text">Show Less</span>
+    </div>
+</ff-asn-group>
+```
+
+**NOTE**: You can change the animation duration by changing the CSS
+`transition-duration` property.
```
torsten-knauf: Is this one the right selector? 

```suggestion
`transition-duration` property of `ff-asn-group .ffw-wrapper`.
```

markdown/3.0/en/api/ff-asn.api.md
omikron-fp: I have that topic on my list. At the moment I don't have enough information to decide if it can be removed safely. Didn't push it any further due to lack of time until release.

markdown/3.0/en/ff-asn.md
omikron-fp: Do you mean `<ff-asn></ff-asn>` by "the example with no effect"? It has the effect that it setups a complete asn with default values/templates.
The custom templates are elaborated on later so I don't see the need to address them here.

markdown/3.0/en/ff-asn.md
omikron-fp: There are several examples involving `ff-asn-group`. Even though this very example has no effect, this step is required by all following examples. It is the entry point if you will. Kind of to convey the basic mechanics of adding a group.
I guess the phrase "has no effect" will make the reader naturally keep on reading.
But maybe I found a suitable wording.

omikron-fp:
```suggestion
To customize filter groups you add `ff-asn-group` elements. However, adding an empty `ff-asn-group` as in the example above has no effect. You will likely want to specify your own HTML template as shown in the following sections.
```

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
...
 <ff-asn-group-slider></ff-asn-group-slider>
 ```
-**NOTE**: The `[for-group]` attribute has a higher priority than the
-`[filter-style]` attribute. This allows overriding certain groups, even
-if you have a default template for that `[filter-style]`
 
-A slider can have a `[data-container="groupCaption"]` attribute but
-instead of detailedLinks and hiddenLinks the slider group accepts a
-`ff-slider-control` and/or an `ff-slider` element.
+A slider can have a `[slot="groupCaption"]` attribute but instead of detailedLinks and hiddenLinks the slider group accepts a `ff-slider-control` and/or an `ff-slider` element.
```
torsten-knauf: I think depending on how you pronounce _f_ you can use _a_ or _an_. For consistency we could decide now how we want to pronounce _ff_ :P

markdown/3.0/en/ff-asn.md
```diff
@@ -1,224 +1,296 @@
...
-
-        <ff-slider step-size="1" style="width: 150px;">
-            <div data-slider="1" class="sliderBtn"></div>
-            <div data-slider="2" class="sliderBtn"></div>
-            <!-- Version < 1.2.2
-            <div id="slider1" class="sliderBtn"></div>
-            <div id="slider2" class="sliderBtn"></div>
-            -->
+        <ff-slider step-size="1" submit-on-release="true">
+            <div slot="slider1" class="sliderBtn"></div>
```
torsten-knauf: Have I already said, that I love the removal of those `id`s? :)

markdown/3.0/en/ff-asn.md
torsten-knauf: I meant line 10 ff (as I should have written initial^^)

markdown/3.0/en/ff-asn.md
torsten-knauf: mh, I would be happier with removing this example _without effect_ and just state that you wrap your custom templates within `ff-asn-group` as suggested above. But I think your solution is suitable. So feel free to decide

markdown/3.0/en/ff-asn.md
omikron-fp:
```suggestion
A slider can have a `[slot="groupCaption"]` attribute but instead of detailedLinks and hiddenLinks the slider group accepts an `ff-slider-control` and/or an `ff-slider` element.
```

PR #171, opened 2019-03-13
add ff-product-teaser-campaign-processor to ff-middleware

omikron-fp:
Previously there was only a dedicated page.
Middleware page is intended to list all available modules therefore this addition.

### DIFF
--- a/markdown/1.x/en/api/ff-middleware.api.md
+++ b/markdown/1.x/en/api/ff-middleware.api.md
@@ -12,3 +12,8 @@ ___
 | **keep-original-in-property**&nbsp;(String) (_optional_)| The name of the property you want to store the unparsed value in. This may be particularly useful when you overwrite the original property but still require the original value. |
 | **entry-separator**&nbsp;(String) (_optional_, default: "&#0448;")| The character configured in FACT-Finder delimiting key-value pairs.  |
 | **key-value-separator**&nbsp;(String) (_optional_, default: "=")| The character configured in FACT-Finder to separate keys from values. |
+
+## `ff-product-teaser-campaign-processor`
+---
+### Properties
+None.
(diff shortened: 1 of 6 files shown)


markdown/1.2/en/ff-middleware.md
```diff
@@ -125,3 +125,27 @@ window.addEventListener("ffReady", () => {
 ```
 
 See the [API tab](/api/1.2/ff-middleware#tab=api) for more details on configuration options.
+
+#### ProductTeaserCampaignProcessor
+
+Starting with FACT-Finder version 7.3 and FACT-Finder Web Components version 1.2.16 _feedback text campaigns_ can be configured as _teaser_.
```
torsten-knauf: I wonder if we should be more precise about the version. A _7.3.5-42_ is needed

markdown/1.2/en/ff-middleware.md
```diff
@@ -125,3 +125,27 @@ window.addEventListener("ffReady", () => {
 ```
 
 See the [API tab](/api/1.2/ff-middleware#tab=api) for more details on configuration options.
+
```
torsten-knauf: I wonder if we should replace the text from api/1.2/ff-product-teaser-campaign-processor through a link to the middleware? Or remove it from the sidebar and add a section to `Campaign` with only a link to the middleware here?

I am not too happy with having almost the exact same text and same feature description in two places written down separately

omikron-fp: Grouping all campaigns in some way would be useful I think. It wouldn't bloat the left navigation as much.
I'll create a separate issue for that. For now it just has to be _good enough_ for the next release.

markdown/1.2/en/ff-middleware.md
omikron-fp: If there is no doubt it is indeed that version, I'll add it quickly.

PR #175, opened 2019-03-14
#174 demo targets

### LINKED ISSUE #173: Change versions to 1.x and 3.x
#### Current behaviour

The version dropdown's items are hardcoded to `1.2` and `3.0`.

#### Expected Behaviour

Minor increments are to be expected.
Either hardcode versions to `1.x` and `3.x` or `v1` and `v3`.
Combinations are thinkable, especially as the term `1.2` seems to be quite established.
e.g. `1.2` and `3.x`

Also adjust the download page's version detection if necessary.

### LINKED ISSUE #174: Target iframes in demo tabs to correct version
#### Current behaviour

All demos draw from an outdated source. (`current`)

#### Expected Behaviour

Demo's `iframe` targets either `1.x` or `3.x` according to selected version.

torsten-knauf:
closes #174
closes #173


src/data/api.js
```diff
@@ -1,9 +1,9 @@
-import api30 from './3.0/api'
-import api12 from './1.2/api'
+import api30 from './3.x/api'
+import api12 from './1.x/api'
```
omikron-fp: We could even go all the way and rename `api30` -> `api3x` and `api12` -> `api1x`

src/data/guides.js
```diff
@@ -1,9 +1,9 @@
-import guides30 from './3.0/guides'
-import guides12 from './1.2/guides'
+import guides30 from './3.x/guides'
+import guides12 from './1.x/guides'
```
omikron-fp: Here too `guides30` -> `guides3x`

PR #186, opened 2019-05-15
Ffweb 1205 currency encoding 1x

torsten-knauf:


markdown/1.x/en/documentation/currency-guide.md
```diff
@@ -105,10 +101,8 @@ Specify the maximum number of digits for all prices.
...
 
 ---
-#### **add-unit-to-filter-groups (Boolean)**
-If present, the unit formatting takes place for all filter groups and it's filters. Note that this is a `Boolean` attribute and is therefore ignoring its value. Read more about [attributes](todo) 
-
-**Example Usage**
-`<ff-communication add-unit-to-filter-groups>`
+#### **add-unit-to-filter-groups (Boolean) (deprecated since 1.3.1)**
+~~If present, the unit formatting takes place for all filter groups and its filters.~~
+Since 1.2.2 the formatting takes places for all filter groups and its filters internally for every response. Setting the attribute was a noop or could lead to false encoding in case of the currency value differs on client and FACT-Finder site. Since 1.3.1 this attribute is always a noop.
```
wdutkiewicz-obj: I am wondering if we should use 'noop' word in official documentation. See [NOP definition](https://en.wikipedia.org/wiki/NOP_(code)). Maybe something like "Since 1.3.1 this attribute is not supported and setting it has no effect.". This comment applies to all 'noop' usages.

omikron-fp: `has no effect` could be a safer construct.

markdown/1.x/en/documentation/currency-guide.md
```diff
@@ -91,24 +87,22 @@ If you are importing multiple price fields like discount prices or something sim
...
```
omikron-fp: Got some grammatical issues.
```
Since 1.2.2, for every response, the formatting is applied internally to all filter groups and their filters. [...] Since 1.3.1 this attribute has no effect.
```
Not sure about how to best express the middle part. Maybe something like:
```
[...] There was a bug that caused the attribute to either have no effect at all or cause wrong encoding under certain circumstances. [...]
```

markdown/1.x/en/documentation/currency-guide.md
```diff
@@ -103,6 +103,6 @@ Specify the maximum number of digits for all prices.
 ---
 #### **add-unit-to-filter-groups (Boolean) (deprecated since 1.3.1)**
 ~~If present, the unit formatting takes place for all filter groups and its filters.~~
-Since 1.2.2 the formatting takes places for all filter groups and its filters internally for every response. Setting the attribute was a noop or could lead to false encoding in case of the currency value differs on client and FACT-Finder site. Since 1.3.1 this attribute is always a noop.
+Since 1.2.2, for every response, the formatting is applied internally to all filter groups and their filters. Hence this attribute had no effect anymore or could lead to incorrect duplicative encoding. Since 1.3.1 this attribute has always no effect.
```
omikron-fp: It's the setting of the attribute that caused the undesired effects, isn't it?
There were two possible outcomes, so an `either...or` clause it adequate.
```
Hence setting this attribute either had no effect at all or it caused incorrect encoding.
```
"...no effect **at all**" emphasizes the no-op behaviour.
The fact that it was _duplicate_ encoding is irrelevant detail.

"always no effect......"

-> `this attribute...`
- `...has no effect`
- `...no longer has any effect`
- `...is ignored`

torsten-knauf: > The fact that it was duplicate encoding is irrelevant detail.

is it? In case someone noticed the duplicated currency symbol it is easier to map the changelog entry for him with this information. Does it hurt to leave it there?

omikron-fp: When I read "duplicate encoding", I think of this `encodeURIComponent(encodeURIComponent(str))`. Nothing currency related.
When you say "encode", do you actually mean "adding a currency symbol"?

torsten-knauf: I would personally map it. But obvious not clear, so let's skip the information

torsten-knauf: done

feel free to _insert a suggestion_ directly next time. I totally love this new github feature , as it shortens feedback loop 👍

PR #189, opened 2019-05-20
FFWEB-1205 deprecate / remove `ff-communication`s attribute `add-unit…

torsten-knauf:
…-to-filter-groups` for 3.x


torsten-knauf:
will update according to https://github.com/FACT-Finder-Web-Components/documentation/pull/186

PR #199, opened 2019-07-02
FFWEB-1254 Update CSS mixins in 3.0 migration guide

wdutkiewicz-obj:
Documentation update according to CSS changes in WebComponents code.

### DIFF
--- a/markdown/3.x/en/documentation/upgrade-guide.md
+++ b/markdown/3.x/en/documentation/upgrade-guide.md
@@ -400,6 +400,9 @@ ffw-transformZero
 
 transformX
 ffw-transformX
+
+noTransition
+ffw-noTransition
 ```
 
 ### ff-sortbox-item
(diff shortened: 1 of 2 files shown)


markdown/3.x/en/documentation/upgrade-guide.md
```diff
@@ -447,7 +468,33 @@ tagCloudLink
 ffw-tagCloudLink
 ```
 
+
+### Replacement of identifiers
```
torsten-knauf:
```suggestion
### Replacement of HTML `id` properties
```

I think _HTML id properties_ is more clear than _identifiers_, isn't it? (same would apply to next comment in the change)

torsten-knauf: I am unsure if we need this section. Weren't those elements meant to be hidden by shadow DOM in the past and hence private? If we don't need it, I think we shouldn't bloat this already huge section.

wdutkiewicz-obj: You are right, we don't need it. I hadn't realized the ids were encapsulated in the Shadow DOM until I saw your comment. I checked for version 3.0.0 and yes - they're all hidden from the user.

wdutkiewicz-obj: This is not relevant any more, as I removed entire section due to your comment.

wdutkiewicz-obj:
@torsten-knauf : please, re-review. I renamed upgrade guide as well to reflect its current state.

PR #200, opened 2019-07-03
add impr link

toarm:


toarm:
> Don't you think it would be nice to stay on our site / have our header in place?
> 
> Maybe spent a new page and embed the link in an iframe.
No, thats how the Impressum links work everywhere and in addition I dont like to spent more time on that topic.

PR #202, opened 2019-08-02
FFWEB-1357 Removed "98% time-saving" references

wdutkiewicz-obj:


src/views/home-view.js
```diff
@@ -170,21 +160,6 @@ class HomeView extends ViewMixin(ReduxMixin(PolymerElement)) {
         <img src="../../images/index/integration-gif.gif">
     </div>
 
-    <hr>
-    <div id="fast" class="row">
-        <div class="row-item">
-            <img src="../../images/index/time_saving_r.gif">
-        </div>
-        <div class="row-item">
```
omikron-fp: I wonder if the text block could remain as it's only the gif that contains the _98%_. Then we wouldn't lose a sales argument.
@toarm, you're the initiator. I guess only you can answer this.

src/views/home-view.js
```diff
@@ -123,16 +123,6 @@ class HomeView extends ViewMixin(ReduxMixin(PolymerElement)) {
...
-    <div class="row">
-        <div class="row-item">
-            <h1><b>Up to 98% faster integration</b></h1>
-            <p style="text-align: center">
-                With the new FACT-Finder Web Components
-            </p>
-        </div>
-    </div>
-
-    <hr>
```
omikron-fp: Removing the `<hr>` makes the tiles from below touch the main image. This looks a little awkward, don't you agree?
Keeping the `<hr>` might look a little odd, too, but placing a margin of `30px` (double of `.container`'s left-padding) between the image and `.container` looks quite pleasing.
Didn't check if `.container` is used somewhere else but it might be best to put the margin on the image's bottom anyway.
What's the opinion of the designer in you?

src/views/home-view.js
wdutkiewicz-obj: I removed it because I thought it looks awkward without the image, but now I can see there is another text block without an image and it looks fine, so I restored the text.

src/views/home-view.js
```diff
@@ -123,17 +123,8 @@ class HomeView extends ViewMixin(ReduxMixin(PolymerElement)) {
      src="../../images/index/hero/hero_image_web_components@2600.png" alt="Banner Image">
 
 <div class="container">
-    <div class="row">
-        <div class="row-item">
-            <h1><b>Up to 98% faster integration</b></h1>
```
omikron-fp: Just for the record:
Please replace text with _"FACT-Finder Web Components"_

src/views/home-view.js
```diff
@@ -123,17 +123,8 @@ class HomeView extends ViewMixin(ReduxMixin(PolymerElement)) {
...
-            <p style="text-align: center">
-                With the new FACT-Finder Web Components
-            </p>
-        </div>
-    </div>
 
-    <hr>
-    <div class="link-box-container">
+    <div class="row link-box-container">
```
omikron-fp: For the record:
`.row` seems to remove the padding between the tiles. Please find a way to make it look the same as before.

PR #207, opened 2019-09-06
FFWEB-1376 Github redirect implementation

wdutkiewicz-obj:
1. Version check removed.
2. Bundle information displayed both for 1.x and 3.x.
3. Removed code unused after the changes. In case of implementing feature selection again, the code can be restored from previous commits.

### DIFF
--- a/src/views/download-view.js
+++ b/src/views/download-view.js
@@ -164,7 +164,7 @@ class DownloadView extends ReduxMixin(PolymerElement) {
                 </p>
             </div>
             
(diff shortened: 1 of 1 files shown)


src/views/download-view.js
```diff
@@ -164,7 +164,7 @@ class DownloadView extends ReduxMixin(PolymerElement) {
                 </p>
             </div>
             
-            <template is="dom-if" if="[[isSelectedVersion3XX]]">
```
omikron-fp: By removing this part, the message box seems to be no longer displayed.
It should be visible.

PR #212, opened 2019-10-10
FFWEB-1355 routing documentation guide follow up

KrzysztofKozlowskiOBSS:


markdown/3.x/en/documentation/routing.md
```diff
@@ -35,6 +35,14 @@ function isSearchPage() {
 }
 ```
 
+#### NOTE
+
+Please do not use `search-immediate` on non-search pages or you will be immediately redirected to the search page.
+
+#### `before-search` event
+
+`before-search` event that is emitted by `ff-searchbox` can be also used to detect certain search events in order to perform redirects. More specifically, this event will occur for all searches that involve the searchbox, but will not occur for ASN, paging, sorting etc. requests. This might come in handy on category pages where entering a new query should trigger a new search that is not be restricted by the current category filter.
```
omikron-fp: Will there be an example of what _"detecting certain search events"_ might look like?

omikron-fp: I'm a little confused.
Is _"entering a new query"_ in `ff-searchbox` ever constrained by the current category filter?
Are there searches involving `ff-searchbox` that are not issued by it?

Mentioning the event's place of occurrence in the same sentence as places where it doesn't kind of creates a connection where there isn't one. If it is only emitted by `ff-searchbox`, how could it possibly occur on `ff-asn`? I feel emphasizing this distinction is more effective when phrased as a separate statement.
I'm led to wonder whether there is more to the `before-search` event that isn't outlined here.

omikron-fp: `is not be` sounds not quite right. Is there a word missing or too much?

KrzysztofKozlowskiOBSS: I wasn't going to provide one.

KrzysztofKozlowskiOBSS: What's wrong with it?

KrzysztofKozlowskiOBSS: > Is "entering a new query" in ff-searchbox ever constrained by the current category filter?

Yes, it can be. We are talking strictly about category pages (with different URL than regular search page) where category filters are "hardcoded" via `add-params`.

> Are there searches involving ff-searchbox that are not issued by it?

No, I never said there are.

> Mentioning the event's place of occurrence in the same sentence as places where it doesn't kind of creates a connection where there isn't one. If it is only emitted by ff-searchbox, how could it possibly occur on ff-asn? 

So what's your suggestion? Do you think that's too obvious that listening to this event will not respond to ASN, paging, sorting as opposed to the beforeDispatchingCallback method from above and I should remove this sentence?

> I'm led to wonder whether there is more to the before-search event that isn't outlined here.

Do you have anything particular in mind?

omikron-fp: Would have expected something like.
`that is not restricted`
`that shall not be restricted`
`that is not to be restricted`

KrzysztofKozlowskiOBSS: Ah, right, I have no clue how `be` got there :D

omikron-fp: Please do. TLDR style is fine. Or a link to a resource to give users some inspiration.
I'm imagining users who are not that familiar with the library but would like to detect search events.

KrzysztofKozlowskiOBSS: That's a regular event that can be listened with `ffSearchBoxElem.addEventListener()`. I can't imagine user that want to use such advanced feature and can't use JS event listeners. But if you really think it's necessary I can add one.

omikron-fp: My first association with _"current category filter"_ is the one manually selected in the ASN. Of course this doesn't constrain the searchbox. If it is `add-params` that leads to such constraints, it might be worth being more precise here by mentioning `add-params`.

> > Are there searches involving ff-searchbox that are not issued by it?
> 
> No, I never said there are.

_"involving"_ is a rather wide term and leaves the impression that there could be. Specifically mentioning the _issuing_ reduces ambiguity IMO.

> So what's your suggestion? Do you think that's too obvious that listening to this event will not respond to ASN, paging, sorting as opposed to the beforeDispatchingCallback method from above and I should remove this sentence?

Even though the concept of components raising events is not unique to FACT-Finder WebComponents and shouldn't require a reminder, experience shows different, So mentioning this distinction has indeed value and is worth keeping. For the sake of eliminating unwanted assumptions about cross-component effects between searchbox and other elements, I see it having more impact when set apart. At the core it's:
Statement: _"searchbox emits before-search only when searchbox issues a request"_
Reminder: _"other request issuing elements (asn etc) are unrelated and logically do not cause searchbox to emit its before-search event."_

> > I'm led to wonder whether there is more to the before-search event that isn't outlined here.
> 
> Do you have anything particular in mind?

No, but that's the issue. But I think this would go away if we can address the things above.

omikron-fp: I'm not thinking about `ffSearchBoxElem.addEventListener("before-search", fn)`, if that is what you are referring to.
I'm more after the search events that can be detected by using `before-search`.
Currently I am stuck at

 _"When I use `before-search`"_,
- _"**how** do I detect certain search events"_
- _"**what** certain search events are available?"_

KrzysztofKozlowskiOBSS: By `certain search events` I mean all the search events that come from `ff-searchbox`
>`before-search` event [..] can be also used to detect **certain search events** in order to perform redirects. **More specifically**, this event will occur for **all searches that involve the searchbox**, but will not occur for ASN, paging, sorting etc. requests. So the fact that this event was fired makes the search event "certain". Do you think that's not clear enough?

KrzysztofKozlowskiOBSS: > My first association with "current category filter" is the one manually selected in the ASN. Of course this doesn't constrain the searchbox. If it is add-params that leads to such constraints, it might be worth being more precise here by mentioning add-params.

Would *"current category filter"* => *"current category page"* work? There might be different ways to implement category pages than using `add-params` I suppose.

> "involving" is a rather wide term and leaves the impression that there could be. Specifically mentioning the issuing reduces ambiguity IMO.

Oh this is what you meant.  I thought I wrote something illogical. Could you use github suggestions if the issue is only about the choice of words?

> Even though the concept of components raising events is not unique to FACT-Finder WebComponents and shouldn't require a reminder, experience shows different, So mentioning this distinction has indeed value and is worth keeping. For the sake of eliminating unwanted assumptions about cross-component effects between searchbox and other elements, I see it having more impact when set apart. At the core it's:
Statement: "searchbox emits before-search only when searchbox issues a request"
Reminder: "other request issuing elements (asn etc) are unrelated and logically do not cause searchbox to emit its before-search event."

I don't understand why do you think that the current version indicates the existence of any cross-component relations. For me the version you suggested means exactly the same and is unnecessarily too verbose.

KrzysztofKozlowskiOBSS: > No, but that's the issue. But I think this would go away if we can address the things above.

In general it was supposed to be rather an honorable mention not a comprehensive documentation of `before-search` event.

omikron-fp: Now I see. Thanks for clarifying! I was genuinely at a loss.
With my current approach of thinking about this a slightly different order and the use of certain (^^) keywords would be clearer:
1) `before-search` is emitted by `ff-searchbox` whenever `ff-searchbox` issues request
2) in `before-search` listener you have the opportunity to respond to only certain/selected/required search requests (for redirects or whatev)
alternatively, _"you can choose if you respond to the event or not"_
3) asn etc are not related, don't trigger the event and so on

This could be followed by a super short example demonstrating the "event filtering". Mostly to give an introduction to what the event object looks like.
```js
ffSearchBoxElem.addEventListener("before-search", e => {
  if (e.whatever) {
    redirect();
  }
});
```

omikron-fp: > Would "current category filter" => "current category page" work?

I very much think so. Maybe even with the addition of _"current cat page's constraints"_. Would that make sense?

KrzysztofKozlowskiOBSS: 👌

omikron-fp: > why do you think that the current version indicates the existence of any cross-component relations

One is the usage of "involves". The other is the continuation of the sentence with _", but will not occur for..."_. It raises the question of why would you mention standard behaviour when there is nothing special about it? Mentioning it in the same sentence gives it a certain urgency. It made me doubt there is nothing special going on and I started wondering if somewhere in WebComponents there are cross-component effects or something else.

omikron-fp: > it was supposed to be rather an honorable mention not a comprehensive documentation of `before-search` event.

We can go this way, too. In that case a link to the comprehensive documentation would be good.
Looking at the current documentation of `before-search` it looks far from comprehensive though. We could address that in a separate issue or place it here now and keep the docs page rather general.

markdown/3.x/en/documentation/routing.md
```diff
@@ -35,6 +35,21 @@ function isSearchPage() {
 }
 ```
 
+#### NOTE
+
+Please do not use `search-immediate` on non-search pages or you will be immediately redirected to the search page.
+
+#### `before-search` event
+
+`before-search` event is emitted by `ff-searchbox` whenever `ff-searchbox` issues a request. In its listener, you have the opportunity to respond to only selected search requests, as asn, paging, sorting etc. will not trigger this event.
```
omikron-fp: If we could switch to all caps ASN, we are good to go.
```suggestion
`before-search` event is emitted by `ff-searchbox` whenever `ff-searchbox` issues a request. In its listener, you have the opportunity to respond to only selected search requests, as ASN, paging, sorting etc. will not trigger this event.
```

omikron-fp: Didn't like the caps here?
```suggestion
`before-search` event is emitted by `ff-searchbox` whenever `ff-searchbox` issues a request. In its listener, you have the opportunity to respond to only selected search requests, as ASN, paging, sorting etc. will not trigger this event.
```

PR #217, opened 2019-10-30
FFWEB-1428: EventAggregator links 404 

a-laurowski:
_addScrollNavigation is only executed in ConnectedCallback when we are initially landed on the EventAggregator subpage. If not, the scroll listeners won't be attached at all.
This Fix adds _addScrollNavigation execution to page change observer

### DIFF
--- a/src/views/api-view.js
+++ b/src/views/api-view.js
@@ -230,6 +230,10 @@ class ApiView extends ViewMixin(ReduxMixin(PolymerElement)) {
             return;
         }
 
+        if (this.subpage === `core-event-aggregator`) {
+            this._addScrollNavigation();
+        }
+
         const fileName = this.subpage;
 
         this.filePath = `markdown/${this.version}/${this.language}/${fileName}.md`;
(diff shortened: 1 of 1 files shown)


src/views/api-view.js
```diff
@@ -230,6 +230,10 @@ class ApiView extends ViewMixin(ReduxMixin(PolymerElement)) {
             return;
         }
 
+        if (this.subpage === `core-event-aggregator`) {
```
omikron-fp: I found that placing this here adds another `marked-render-complete` handler every time the EventAggregator page is visited.
A more general and extensible solution seems to be simply removing the condition at line 197. Although `_addScrollNavigation` would need some adjustment to only attach a `click` handler when an anchor is of this "scroll-to" kind. At the moment this should be easily distinguishable by the `href` formatting.

Although adding options `{ once: true }` to the `marked-render-complete` handler seems to work fine, too, it appears to be a rather hacky quickfix.

a-laurowski: I removed checking if current subpage is an 'core-event-aggregator' and modified connectedCallback. I'm not be sure if filtering anchors which should trigger the scroll is fine. For now I'm distinguihs them by checking href attribute value which in case of scrolling-anchors doesn't start with the '/'

PR #221, opened 2019-11-05
FFWEB-1370 add searchable ASN documentation

KrzysztofKozlowskiOBSS:


markdown/3.x/en/ff-asn.md
```diff
@@ -302,3 +302,50 @@ The default value is `false`.
     Reset Filters
 </ff-asn-remove-all-filter>
 ```
+
+## Searchable ASN groups
+
+When the number of available filters in the ASN group is large you may want to render an additional search field to allow user to narrow them.
```
omikron-fp: How do you feel about this?
```suggestion
When the number of available filters in the ASN group is large, you may want to render a search field inside the group to allow the user to narrow down displayed filters.
```

markdown/3.x/en/ff-asn.md
```diff
@@ -302,3 +302,50 @@ The default value is `false`.
...
+
+To enable this feature you have to set the minimum number of filters required to render the search field in a given group. This is done by setting the `searchable-from` attribute on `ff-asn` element.
```
omikron-fp:
```suggestion
To enable this feature you have to set the minimum number of filters required to render the search field in a given group. This is done by setting the `searchable-from` attribute on the `ff-asn` element.
```

markdown/3.x/en/ff-asn.md
```diff
@@ -302,3 +302,50 @@ The default value is `false`.
...
+If you wish to exclude some groups from being searchable regardless of the number of filters, you can set the `non-searchable` attribute on such `ff-asn-group` element.
```
omikron-fp:
```suggestion
If you wish to exclude some groups from being searchable regardless of the number of filters, you can set the `non-searchable` attribute on such an `ff-asn-group` element.
```

markdown/3.x/en/ff-asn.md
```diff
@@ -302,3 +302,50 @@ The default value is `false`.
...
+
+### Search field template
+
+The default search field template is `<div slot="filterSearch"><input></div>`, however it can be easily changed by providing own content for `filterSearch` slot in `ff-asn-group` template.
```
omikron-fp: I feel the default template has quite some importance that justifies putting it in its own HTML block. Kind of like so to increase rediscoverability.
```suggestion
The default search field template is
```html
<div slot="filterSearch"><input></div>
    ```
However it can be changed by providing own content for `filterSearch` slot in `ff-asn-group` template. 
```

Let's also avoid words like "easy." Every reader feels different about this. We'll just show them how to do it and let them decide whether it's easy or not.

PR #225, opened 2019-11-25
FFWEB-1458 Added more details about 'unresolved' attribute and its usage

wdutkiewicz-obj:


markdown/3.x/en/documentation/include-scripts.md
```diff
@@ -24,16 +24,19 @@ Depending on the browser's page loading time you might encounter ugly flashes of
 ```html
 <ff-record-list unresolved></ff-record-list>
 ```
-Also add the following CSS rule to the top of the page before the appearance of FACT-Finder Web Components:
+Also add suitable CSS rule to the top of the page before the appearance of FACT-Finder Web Components.
+
+The following piece of CSS hides element marked as unresolved (as well as its descendants) in the document:
```
a-laurowski:
```suggestion
The following CSS snippet hides element marked as unresolved (as well as its descendants) in the document:
```

markdown/3.x/en/documentation/include-scripts.md
```diff
@@ -24,16 +24,19 @@ Depending on the browser's page loading time you might encounter ugly flashes of
...
         [unresolved] {
-            visibility: hidden;
+            display: none;
         }
     </style>
 </head>
 ```
 
-**Note:** The `<style>` tag must be inlined in the main document. If you add it to a CSS file, it might not be loaded fast enough and FOUC might still occur.
+**NOTE:** There are several ways to hide HTML element, like setting its `opacity` or `visibility` style. Unlike `display: none`, elements with `opacity: 0` or `visibility: hidden` will occupy their place in the document, but will not be visible. You can apply any style you need to `unresolved` elements. Please, see the official CSS specification for the details.
```
a-laurowski:
```suggestion
**NOTE:** There are several ways to hide HTML element, like setting its `opacity` or `visibility` property. Unlike `display: none`, elements with `opacity: 0` or `visibility: hidden` will occupy their place in the document, but will not be visible. You can apply any style you need to `unresolved` elements. Please, see the official CSS specification for the details.
```

markdown/3.x/en/documentation/include-scripts.md
```diff
@@ -24,16 +24,19 @@ Depending on the browser's page loading time you might encounter ugly flashes of
 ```html
 <ff-record-list unresolved></ff-record-list>
 ```
-Also add the following CSS rule to the top of the page before the appearance of FACT-Finder Web Components:
+Also add suitable CSS rule to the top of the page before the appearance of FACT-Finder Web Components.
```
omikron-fp:
```suggestion
Also add suitable CSS rules to the top of the page before FACT-Finder Web Components are referenced.
```
_"the appearance of WebComponents"_ sounds confusing in the context of elements' visibility. I think we should take the opportunity and improve this part. But feel free to tweak it if you can think of something better.

markdown/3.x/en/documentation/include-scripts.md
```diff
@@ -24,16 +24,19 @@ Depending on the browser's page loading time you might encounter ugly flashes of
...
+
+The following CSS snippet hides element marked as unresolved (as well as its descendants) in the document:
 
 ```html
 <head>
     <style>
         [unresolved] {
-            visibility: hidden;
+            display: none;
```
omikron-fp: I'm not sure whether `visibility: hidden` isn't the generally more suitable style rule. `display: none` should only be useful for elements that don't usually take up space in the regular document flow. Like ff-suggest's popup.

I would like to see a second rule here. Like:
```html
ff-suggest[unresolved] {
    display: none;
}
```
So it becomes clear that it might make sense to distinguish methods of hiding unresolved elements.

markdown/3.x/en/documentation/include-scripts.md
```diff
@@ -24,16 +24,19 @@ Depending on the browser's page loading time you might encounter ugly flashes of
...
         }
     </style>
 </head>
 ```
 
-**Note:** The `<style>` tag must be inlined in the main document. If you add it to a CSS file, it might not be loaded fast enough and FOUC might still occur.
+**NOTE:** There are several ways to hide HTML element, like setting its `opacity` or `visibility` CSS property. Unlike `display: none`, elements with `opacity: 0` or `visibility: hidden` will occupy their place in the document, but will not be visible. You can apply any style you need to `unresolved` elements. Please, see the official CSS specification for the details.
```
omikron-fp:
```suggestion
**NOTE:** There are several ways to hide an HTML element, like setting its `opacity` or `visibility` CSS property. Unlike `display: none`, elements with `opacity: 0` or `visibility: hidden` will occupy their place in the document, but will not be visible. You can apply any style you need to `unresolved` elements. Please, see the official CSS specification for the details.
```

markdown/3.x/en/documentation/include-scripts.md
```diff
@@ -24,16 +24,19 @@ Depending on the browser's page loading time you might encounter ugly flashes of
...
+The `<style>` tag must be inlined in the main document. If you add it to a CSS file, it might not be loaded fast enough and FOUC might still occur. `unresolved` attribute is removed when the component is fully initialized and connected to the document.
```
omikron-fp: The behaviour of `resolved` being automatically removed after initialisation is rather important and deserves being mentioned in the main text outside the "note".
Appended to the first paragraph `Depending on the browser's...` seems like a suitable place.

```suggestion
The `<style>` tag must be inlined in the main document. If you add it to a CSS file, it might not be loaded fast enough and FOUC might still occur. The `unresolved` attribute is removed when the component is fully initialized and connected to the document.
```

markdown/3.x/en/documentation/include-scripts.md
wdutkiewicz-obj: I think `display: none` is fine, until you set fixed size of the component (width and height). What is the benefit of `visibility: hidden` if you don't know its size? Let's say we have `ff-asn` with some custom templates. It takes some small place, but when rendered it will expand its area. Having something under filters, you will a blank space over it and then suddenly filters appear. In such case user can add some spinner styles until the component shows up. Can you find some practical use of `visibility: hidden` in the FF Web Components? Besides, they render quickly (event when waiting for data from FACT-Finder they mostly render with empty initial data and appear after retrieving something from the server).

markdown/3.x/en/documentation/include-scripts.md
wdutkiewicz-obj: Please, check now.

PR #226, opened 2019-11-28
FFWEB-1494 extend Angular's FAQ section

KrzysztofKozlowskiOBSS:


markdown/3.x/en/faq.md
```diff
@@ -3,7 +3,21 @@
 #### Question:
 I want to use FACT-Finder Web Components with Angular, but data binding used in the components conflicts with Angular's `{{ }}` binding. Can you fix it?
 #### Answer:
-Default binding delimiters can be changed via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine).
+You can add Angular's `ngNonBindable` attribute to the parent element to prevent Angular from parsing double curly braces. If you have to keep both Angular and FACT-Finder WebComponents bindings within the same element, you can use: 
+```html
+<ng-container ngNonBindable>
+    {{FF Binding}}
+</ng-container>
```
wdutkiewicz-obj: This example is fine, but taken from the context it is not clear. I guess `<ng-container ngNonBindable>` should wrap some FACT-Finder WebComponent with template, like:
```
<div>
    <!-- Any bindings here will be handled by Angular -->
    <ng-container ngNonBindable>
        <ff-template scope="result">
            <!-- This binding will be ignored by Angular and handled by ff-template  -->
            <span>Product found: {{resultCount}}</span>
        </ff-template>
    </ng-container>
</div>```

File: markdown/3.x/en/faq.md [hunk 2]
```diff
@@ -3,7 +3,21 @@
...
+```
+The `<ng-container>` tag will not be rendered in the DOM.
+
+Alternatively, you can change the default FACT-Finder Web Components binding delimiters via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine). Please note that HTML binding will still require a single curly bracket, what will cause an Angular error. You can utilize `ngNonBindable` to handle this issue as well:
```
wdutkiewicz-obj: I'd change `Please note that HTML binding will still require...` -> `Please note that binding HTML still requires...`. "HTML binding" sounds like binding to HTML, whereas "binding HTML" suggests that you bind HTML - and I guess that is the intention, isn't it?

markdown/3.x/en/faq.md
```diff
@@ -3,7 +3,21 @@
...
+```html
+<ng-container ngNonBindable>
+    {{{FF HTML Binding}}}
+</ng-container>
+```
+In this case, please use triple curly braces, no matter the custom delimiters you chose.
```
wdutkiewicz-obj: Do I understand correctly: you change Mustache delimiters to custom ones, like `[` and you still have to use `{{{ }}}` when FF Web Component is nested in `<ng-container ngNonBindable></ng-container>`? Wasn't it that we had to use `[[{}]]` to bind HTML using custom delimiters, or was there a change in a meanwhile?

markdown/3.x/en/faq.md
```diff
@@ -3,7 +3,21 @@
...
+    {{FF Binding}}
+</ng-container>
+```
+The `<ng-container>` tag will not be rendered in the DOM.
+
+Alternatively, you can change the default FACT-Finder Web Components binding delimiters via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine). Please note that HTML binding will still require a single curly bracket, what will cause an Angular error. You can utilize `ngNonBindable` to handle this issue as well:
+```html
+<ng-container ngNonBindable>
+    {{{FF HTML Binding}}}
+</ng-container>
```
wdutkiewicz-obj: The same comment as above - please, give this some more context.

markdown/3.x/en/faq.md
wdutkiewicz-obj: I would add information about FF Web Components version that supports custom Mustache delimiters. The documentation is (or at least should be) for the latest one, we cannot assume that the users update their Web Components with every release.

markdown/3.x/en/faq.md
KrzysztofKozlowskiOBSS: You understand correctly, you have to use `{{{ }}}` in this case. Angular will complain about `[[{ }]]`

markdown/3.x/en/faq.md
KrzysztofKozlowskiOBSS: We don't do it for other features we implement. I don't think we need it here.

markdown/3.x/en/faq.md
KrzysztofKozlowskiOBSS: Actually, it shouldn't :D I'll extend the examples then.

markdown/3.x/en/faq.md
wdutkiewicz-obj: This FAQ answer is valid only for version 3.6.0 or above. I don't expect users with 3.5.0 or below who try this will check CHANGELOG.md to verify if their version supports this feature. They will rather complain and report "not working". If we don't do it for other features it doesn't mean it is fine not to do it.

KrzysztofKozlowskiOBSS: I won't, we should be consistent. If we start putting a note with a version for every single feature we implement the documentation will look terrible.

omikron-fp: We could add another FAQ entry:
Q: Which version is the documentation referring to?
A: Latest

KrzysztofKozlowskiOBSS: @omikron-fp That's actually a great idea.

markdown/3.x/en/faq.md
KrzysztofKozlowskiOBSS: Actually, this example can be used in any context, that's just the way to bind html. You want to bind HTML - you do `<whatever-tag ngNonBindable>{{{property}}}</whatever-tag>`

markdown/3.x/en/faq.md
```diff
@@ -1,9 +1,32 @@
...
+<ff-record> 
+    {{Angular Binding}}
+    <ng-container ngNonBindable>
+        {{FF Binding}}
+    </ng-container>
+</ff-record>
+```
+The `<ng-container>` tag will not be rendered in the DOM.
+
+Alternatively, you can change the default FACT-Finder Web Components binding delimiters via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine). Please note that binding HTML still requires a single curly bracket, what will cause an Angular error. You can utilize `ngNonBindable` to handle this issue as well:
```
omikron-fp: When quickreading, I tripped over _"... binding HTML requires a single curly ..."_. Sounded a bit to me as if binding HTML would overwrite other ways of data binding. I find _"additional"_ helps understanding the workings even when you are not fully aware of the context.
Regarding _"single curly"_ vs _"curly"_, I'm rather indifferent. Both are fine.
Otherwise, looks good to me.
```suggestion
Alternatively, you can change the default FACT-Finder Web Components binding delimiters via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine). Please note that binding HTML still requires an additional pair of (single) curly brackets, what will cause an Angular error. You can utilize `ngNonBindable` to handle this issue as well:
```

markdown/3.x/en/faq.md
```diff
@@ -20,7 +20,7 @@ You can add Angular's `ngNonBindable` attribute to the parent element to prevent
 ```
 The `<ng-container>` tag will not be rendered in the DOM.
 
-Alternatively, you can change the default FACT-Finder Web Components binding delimiters via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine). Please note that binding HTML still requires a single curly bracket, what will cause an Angular error. You can utilize `ngNonBindable` to handle this issue as well:
+Alternatively, you can change the default FACT-Finder Web Components binding delimiters via `ff-communication`'s `mustache-delimiters` attribute. For more information please check the `Underlying Engine (Mustache)` section of the [Template Engine documentation](/documentation/3.x/template-engine). Please note that binding HTML still requires an additional pair of (single) curly brackets, what will cause an Angular error. You can utilize `ngNonBindable` to handle this issue as well:
```
omikron-fp: Oops, I meant the `(single)` to be replaced by `single` or be removed altogether. I guess have to be more precise about what I actually mean.
Both work fine for me. You may choose.

KrzysztofKozlowskiOBSS: I just blindly committed your suggestion :D fixed

wdutkiewicz-obj:
> Any concerns left @wdutkiewicz-obj?

I think I we can discuss it offline (on-site) with Krzysztof. Regarding this PR - as long as it is understandable for potential FF WebComponents user - it's OK.

KrzysztofKozlowskiOBSS:
> > Any concerns left @wdutkiewicz-obj?
> 
> I think I we can discuss it offline (on-site) with Krzysztof. Regarding this PR - as long as it is understandable for potential FF WebComponents user - it's OK.

I'm not going to be in the office today, so let's keep the discussion here.

omikron-fp:
@wdutkiewicz-obj, you are still on "changes requested." Anything left to discuss?

PR #227, opened 2019-12-04
Ffweb 1454 document keep category path

KrzysztofKozlowskiOBSS:


markdown/3.x/en/api/ff-asn.api.md
```diff
@@ -98,3 +98,12 @@ ___
...
 | **submit-on-input**&nbsp;(String) **Options**&nbsp;"true",&nbsp;"false" (default: "false") | If this attribute is set to true, the input fields accept values. If set to false, you cannot focus the input fields. |
 | **unit**&nbsp;(String) | The unit of measurement. E.g. € |
+
+## `ff-asn-remove-all-filter`
+___
+### Properties
+| Name | Description |
+| ---- | ----------- |
+| **show-always**&nbsp;(Boolean) |  Defines if the element is only visible when at least one removable filter is set, or always. |
+| **remove-params**&nbsp;(Boolean) | Defines whether the resetting request should only contain a search with the current search query or if it should keep the current parameters like `products-per-page` and `sort` and only remove those parameters that were applied by the ASN (recognizable by the `filter*` prefix). |
```
omikron-fp: Had to read it a couple of times to fully understand. I think the description would benefit from shorter sentences.
What do you think about this one?
```suggestion
| **remove-params**&nbsp;(Boolean) | Setting this attribute removes all parameters (except the search query) from the resetting request. When not set, only parameters that were applied by the ASN (recognizable by the `filter*` prefix) are removed. Other parameters like `products-per-page`, `sort` and custom parameters will be unaffected. |
```

KrzysztofKozlowskiOBSS: I simply copy-pasted it from the documentation markdown, but I like your version more.
Should we change the other one as well?

omikron-fp: Totally missed the other location. But yes, let's change it. It needs it :)

PR #229, opened 2019-12-17
FFWEB-1512: `infinite-scroll` NOTE

a-laurowski:
Added `infinite-scroll` NOTE about using inside recommendation or pushed products

### DIFF
--- a/markdown/1.x/en/ff-record-list.md
+++ b/markdown/1.x/en/ff-record-list.md
@@ -129,7 +129,7 @@ Note: When you manually load a page, e.g. from an URL with the 3rd page set in i
 We recommend using higher values for `[infinite-debounce-delay]` and `infinite-scroll-margin` like:
 
 **Note:**
-Please do not use `infinite-scroll` in `ff-record-list` inside components like `ff-recommendation` or `ff-pushed-products`. This attribute is not designed to work with them and using it may cause unwanted side effect
+Please do not use `infinite-scrolling` when the `ff-record-list` is inside a component like `ff-recommendation` or `ff-pushed-products`. The attribute is not designed to work in this context and using it may cause unwanted side effects.
 
 ```html
 <ff-record-list infinite-scrolling
--- a/markdown/1.x/en/ff-record-list.md
+++ b/markdown/1.x/en/ff-record-list.md
@@ -128,9 +128,6 @@ Note: When you manually load a page, e.g. from an URL with the 3rd page set in i
 
 We recommend using higher values for `[infinite-debounce-delay]` and `infinite-scroll-margin` like:
 
-**Note:**
-Please do not use `infinite-scrolling` when the `ff-record-list` is inside a component like `ff-recommendation` or `ff-pushed-products`. The attribute is not designed to work in this context and using it may cause unwanted side effects.
-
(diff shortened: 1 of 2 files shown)


markdown/1.x/en/ff-record-list.md
```diff
@@ -128,6 +128,9 @@ Note: When you manually load a page, e.g. from an URL with the 3rd page set in i
 
 We recommend using higher values for `[infinite-debounce-delay]` and `infinite-scroll-margin` like:
 
+**Note:**
+Please do not use `infinite-scroll` in `ff-record-list` inside components like `ff-recommendation` or `ff-pushed-products`. This attribute is not designed to work with them and using it may cause unwanted side effect
```
omikron-fp: `in ... inside ...` felt a bit bumpy. What's your opinion on this?
```suggestion
Please do not use `infinite-scrolling` when the `ff-record-list` is inside a component like `ff-recommendation` or `ff-pushed-products`. The attribute is not designed to work in this context and using it may cause unwanted side effects.
```

markdown/1.x/en/ff-record-list.md
```diff
@@ -128,6 +128,9 @@ Note: When you manually load a page, e.g. from an URL with the 3rd page set in i
 
 We recommend using higher values for `[infinite-debounce-delay]` and `infinite-scroll-margin` like:
 
+**Note:**
```
omikron-fp: The _Note_ is probably best placed below the code listing. The `...like:` from above doesn't work anymore.

markdown/3.x/en/ff-record-list.md
```diff
@@ -128,6 +128,9 @@ Note: When you manually load a page, e.g. from an URL with the 3rd page set in i
 
 We recommend using higher values for `[infinite-debounce-delay]` and `infinite-scroll-margin` like:
 
+**Note:**
```
omikron-fp: Same remarks as in `1.x` file.

a-laurowski: changed

markdown/3.x/en/ff-record-list.md
```diff
@@ -141,5 +141,7 @@ We recommend using higher values for `[infinite-debounce-delay]` and `infinite-s
     </ff-record>
 </ff-record-list>
 ```
+**Note:**
+Please do not use `infinite-scroll` in `ff-record-list` inside components like `ff-recommendation` or `ff-pushed-products`. This attribute is not designed to work with them and using it may cause unwanted side effect
```
omikron-fp: Please consider applying the textual changes from the `1.x` Notes.

markdown/3.x/en/ff-record-list.md
wdutkiewicz-obj: Please, unify notes so they look the same in 3.x and 1.x - now they differ. I guess 1.x is the correct one.

a-laurowski: done

markdown/3.x/en/ff-record-list.md
a-laurowski: changed

PR #234, opened 2020-01-20
FFWEB-1537 Demo link fix in version 3.x

wdutkiewicz-obj:
Fixed 2 occurrences of incorrect link in version 3.x.


a-laurowski:
> This changes could be also applied in corresponding pages in 1.x. Can You change these files too?

PR #245, opened 2020-03-27
FFWEB-1576 `disable-input-fields` documentation update

wdutkiewicz-obj:


markdown/3.x/en/api/ff-asn.api.md
```diff
@@ -95,7 +95,7 @@ ___
 ### Properties
 | Name | Description |
 | ---- | ----------- |
-| **disable-input-fields**&nbsp;(String) **Options**:&nbsp;"true",&nbsp;"false" (default: "false") | If this attribute is set to false, the input fields accept values. If set to true, you cannot focus the input fields. |
+| **disable-input-fields**&nbsp;(String) **Options**:&nbsp;"true",&nbsp;"false" (default: "false") | If this attribute is set to false, the input fields accept values. If set to true, the fields are disabled by setting `disabled` attribute on them. Therefore they cannot be set focus on. |
```
omikron-fp: There's quite a few occasions of `set`. Could you imagine something like this?
```suggestion
| **disable-input-fields**&nbsp;(String) **Options**:&nbsp;"true",&nbsp;"false" (default: "false") | If this attribute is set to false, the input fields accept values. If set to true, the fields are disabled by setting the `disabled` attribute on them. Then they cannot be focused. |
```

wdutkiewicz-obj: "be focused" was my initial choice.

PR #246, opened 2020-04-02
FFWEB-1623 Add `ff-loading-spinner` documentation

wdutkiewicz-obj:


markdown/3.x/en/ff-loading-spinner.md
```diff
@@ -0,0 +1,83 @@
...
+    const spinner = document.querySelector('ff-loading-spinner');
+    spinner.show();
+});
+</script>
+```
+
+#### Changing color
+
+It is possible to change the element's stroke-color to fit into website design. A CSS color value should be passed to
+`stroke-color` attribute.
```
omikron-fp, line 72: I think it's important here to mention it is **your** design not just any. It might be easier for readers to relate to the documentation.
Furthermore, we can be more specific that passing the color value is mandatory. There is no other official way to do so, so _should_ seems a little too weak here.
```suggestion
It is possible to change the element's stroke-color to fit into your website design. A CSS color value needs to be passed to the `stroke-color` attribute.
```

Alternatively to _"your (website design)"_, i see _"your website's design"_ as a possiblity. It's a little less buzzwordy and would also work well. But I also think this mainly a matter of style. Pick the one you deem more appropriate for the context.
```suggestion
It is possible to change the element's stroke-color to fit into your website's design. A CSS color value needs to be passed to the `stroke-color` attribute.
```

markdown/3.x/en/api/ff-loading-spinner.api.md
```diff
@@ -0,0 +1,14 @@
...
+| ---- | ----------- |
+| **is-active**&nbsp;(Boolean) (Default: "false") | If set, the element is shown and animated. Otherwise the element is hidden. |
+| **stroke-color**&nbsp;(String) (Default: "#4285F4") | Takes a CSS color value and applies it to the spinning stroke. |
+| **manual**&nbsp;(Boolean) (Default: "false") | If set, the element will not subscribe to any default events and subsequently has to be shown or hidden manually | 
+
+### Methods
+| Name | Description |
+| ---- | ----------- |
+| **show** | Shows the element. Calling this method is equivalent to setting the `is-active` attribute. |
+| **hide** | Hides the element. Calling this method is equivalent to removing the `is-active` attribute. |
```
omikron-fp, line 14: Most places list methods with `()`. This is useful to explicitly show which parameters are available.
I know it's inconsistent at the moment, so we should clean up other places, too, some time.
```suggestion
| **show()** | Shows the element. Calling this method is equivalent to setting the `is-active` attribute. |
| **hide()** | Hides the element. Calling this method is equivalent to removing the `is-active` attribute. |
```

omikron-fp:
It might be worth waiting for the [related demo](https://github.com/FACT-Finder-Web-Components/demos/pull/29) before merging this one. We could immediately reference it in the "Demo" tab.

wdutkiewicz-obj:
> It might be worth waiting for the [related demo](https://github.com/FACT-Finder-Web-Components/demos/pull/29) before merging this one. We could immediately reference it in the "Demo" tab.

Sure, I have already activated the tab. Please, review demos PR and we can proceed.

markdown/3.x/en/ff-loading-spinner.md
```diff
@@ -1,88 +1,82 @@
...
 #### Manual mode
 
-In addition to the default 'subscription' mode, the element also has a _manual mode_
-in which the element will _not_ be subscribed to any events. 
-It is activated if the `manual` HTML attribute is present. 
-Adding the attribute will unsubscribe the element form all events. Likewise, re-adding
-the attribute will subscribe the element again.
+In addition to the default subscription mode, the element also has a _manual mode_ in which the element will _not_ be 
+subscribed to any events. Manual mode is activated when the `manual` HTML attribute is present. Adding the attribute will 
+unsubscribe the element form all events. Likewise, removing the attribute will subscribe the element again. Manual mode
```
omikron-fp, line 35: I missed one.
```suggestion
unsubscribe the element from all events. Likewise, removing the attribute will subscribe the element again. Manual mode 
```

omikron-fp:
Related demo is already deployed and I tested the demo tab locally. Works!

PR #247, opened 2020-04-07
FFWEB-1567: introduced `topic` attribute in `ff-asn`

a-laurowski:


markdown/3.x/en/api/ff-asn.api.md
```diff
@@ -25,6 +25,7 @@ ___
 | **select-box**&nbsp;(String) **Options**:&nbsp;"true",&nbsp;"false" (default: "false") | Use this when the hiddenLinks should be an HTML select element |
 | **disable-auto-expand**&nbsp;(Boolean) | Prevents group from being automatically expanded when it contains any active filters. |
 | **not-searchable**&nbsp;(Boolean) (default: false) | Prevents group from being searchable even if it satisfies `searchable-from` condition.  |
+| **topic**&nbsp;(String) | Defines if the component should subscribe to a custom topic instead of the default one. Please note that in this case you need to manually dispatch asn data.  For more information regarding the data's structure, please refer to **Dispatching data manually** in the [ff-asn documentation](/api/3.x/ff-asn#tab=docs). |
```
omikron-fp, line 28: `topic` should be under `ff-asn`. It currently appears to be under `ff-asn-group`.
I also noted that there is no mention of the `subscribe` attribute. Would you add that one, too, if possible?

omikron-fp, line 28: There is a double space before `For more information....`. Is it intended to be a line break?
I think line breaks don't work in tables.

```suggestion
| **topic**&nbsp;(String) | Defines if the component should subscribe to a custom topic instead of the default one. Please note that in this case you need to manually dispatch ASN data. For more information regarding the data's structure, please refer to **Dispatching data manually** in the [ff-asn documentation](/api/3.x/ff-asn#tab=docs). |
```

markdown/3.x/en/ff-asn.md
```diff
@@ -357,3 +357,17 @@ This template has to contain exactly one `input` element.
         </div>
     </ff-asn-group>
 ```
+
+### Dispatching data manually
+If you need to manually dispatch data, for instance when component is subscribed to custom topic, you can use `ResultDispatcher` to achieve that.
```
omikron-fp, line 362:
```suggestion
If you need to manually dispatch data, for instance when the component is subscribed to a custom topic, you can use `ResultDispatcher` to achieve that.
```

markdown/3.x/en/ff-asn.md
```diff
@@ -357,3 +357,17 @@ This template has to contain exactly one `input` element.
...
+ASN component requires `groups` property content of FACT-Finder response to be passed. Any additional preparation is not needed.
```
omikron-fp, line 363:
```suggestion
The ASN component requires the `groups` property's content of the FACT-Finder response to be passed.
No additional preparation is needed. 
```

markdown/3.x/en/ff-asn.md
```diff
@@ -357,3 +357,17 @@ This template has to contain exactly one `input` element.
         </div>
     </ff-asn-group>
 ```
+
+### Dispatching data manually
```
omikron-fp, line 361: I don't see manual data dispatching as part of _Searchable ASN groups_ so I think it's best to increase the headline level.
```suggestion
## Dispatching data manually
```

There is also an indentation error on **line 322** messing up the displayed formatting in GitHub. Would you mind fixing that with the other changes?

a-laurowski: Indentation has been removed

markdown/3.x/en/api/ff-asn.api.md
a-laurowski: It's mistake. I removed the redundant space.

a-laurowski: Fixed.
I added `subscribe` attribute, but also noticed that its type is `string`. in all occurrences. Should I change it to Boolean?

omikron-fp: Didn't check all occurrences but from what I found StringyBooleans are noted as type `String`.
`Boolean` is reserved to true boolean attributes as far as I know.

PR #250, opened 2020-04-23
FFWEB-1629 Remove obsolete references

wdutkiewicz-obj:
Removed `<link rel="import" href="pathToHtmlImport/elements.build.with_dependencies.html">` from code snippets as they are not supported in FACT-Finder WebComponents 3.x.


omikron-fp:
Rebased onto latest `release/3.x` because there was a force-push to a commit prior to this branch which messed up this PRs diff.

PR #253, opened 2020-05-25
FFWEB-1697 Documentation update for `ff-recommendation`

wdutkiewicz-obj:


markdown/3.x/en/ff-recommendation.md
```diff
@@ -26,6 +26,9 @@ This `ff-record-list` does not listen to the default search result and gets its
 </ff-recommendation>
 ```
 
+**IMPORTANT** Although `ff-recommendation` tries to find nested `ff-record-list` and change its `subscribe` property to `false`,
+it should be set explicitly to `false` by user in HTML code to prevent console warnings and avoid race conditions.
```
omikron-fp, line 30: Favour single lines per sentence for super slick diffs.
(Also added an `a` before _"nested"_)
```suggestion
**IMPORTANT** Although `ff-recommendation` tries to find a nested `ff-record-list` and change its `subscribe` property to `false`, it should be set explicitly to `false` by user in HTML code to prevent console warnings and avoid race conditions.
```

PR #254, opened 2020-06-15
Fix typo in user-id parameter (ff-communication API)

fmarangi:


markdown/3.x/en/api/ff-communication.api.md
```diff
@@ -12,7 +12,6 @@ ___
 | **default-query**&nbsp; (String) (default: '*') | Determines which search term should is used by default if no search term provided in as http parameter or in a search event object. |
```
omikron-fp, line 12: Could we fix this one, too, while we're at it?

PR #261, opened 2020-08-07
Add activation notice to documentation

fmarangi:
Unfortunately, we are often becoming tickets from customers who try our library before the Web Components security activation. Service Desk asked therefore for a notice in our documentation.

### DIFF
--- a/markdown/3.x/en/faq.md
+++ b/markdown/3.x/en/faq.md
@@ -122,6 +122,18 @@ If this doesn't seem to solve the issue, please contact us for a custom solution
 ---
 
 ## Errors
+#### Error:
+
+`Access to XMLHttpRequest at 'https://my.fact-finder.de/FACT-Finder/Search.ff?query=*&channel=my_channel' has been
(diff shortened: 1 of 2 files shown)


omikron-fp:
This might also be a good FAQ entry. What do you think about adding one among the upper questions?

markdown/3.x/en/faq.md
```diff
@@ -122,10 +122,24 @@ If this doesn't seem to solve the issue, please contact us for a custom solution
...
 
 ## Errors
+#### Error:
+
+`Access to XMLHttpRequest at 'https://my.fact-finder.de/FACT-Finder/Search.ff?query=*&channel=my_channel' has been
+blocked by CORS policy: Response to preflight request doesn't pass access control check: Redirect is not allowed for a
+preflight request.`
+
+#### Solution:
+In order to use our Web Components library, this must be activated on your FACT-Finder server.
```
omikron-fp, line 132: Could we specify what exactly `this` is? The answer would become standalone and could be copy and pasted without much context.
Otherwise good 👍

PR #268, opened 2020-09-09
FFWEB-1782 for-group references associatedFieldName

KrzysztofKozlowskiOBSS:


markdown/4.x/en/api/ff-asn.api.md
```diff
@@ -21,7 +21,7 @@ ___
 | ---- | ----------- |
 | **opened**&nbsp;(Boolean) | Determines if the filter group is expanded. If you want to have a group opened by default: `<ff-asn-group opened>` does the trick. |
 | **collapsible**&nbsp;(String)&nbsp;**Options**:&nbsp;"true",&nbsp;"false"&nbsp;(default: "true") | Determines, if the filter group is collapsible. |
-| **for-group**&nbsp;(String)&nbsp;(default: "all")| Determines, which filter group the template should be applied to. If set to `"all"`, the template is applied to all groups if no other template matches the requirements. |
+| **for-group**&nbsp;(String)&nbsp;(default: "all")| Determines which filter group the template should be applied to. Its value should correspond to a source field name (associatedFieldName). If set to `"all"`, the template is applied to all groups if no other template matches the requirements. |
```
omikron-fp, line 24: I'm wondering whether a word stronger than `should` would be clearer in this context.
`Its value must correspond...` or `Its value corresponds....`.
And perhaps `to THE source field name` might be more accurate. Not sure if there are more options as sources.

KrzysztofKozlowskiOBSS: I was also thinking about this `should` vs `must`, but `"all"` is a kind of exception here.

omikron-fp: I see. Would swapping the sentences make this work? Kind of like this?
```suggestion
| **for-group**&nbsp;(String)&nbsp;(default: "all")| Determines which filter group the template should be applied to. If set to `"all"`, the template is applied to all groups if no other template matches the requirements. Otherwise, its value corresponds to the source field name (associatedFieldName). |
```

omikron-fp: Let's apply this to the other occurrence, too.

omikron-fp (approved):

KrzysztofKozlowskiOBSS:
squashed fixup commits

omikron-fp (approved):

PR #270, opened 2020-09-14
FFWEB-1803: Publish `dispatchRaw` function

a-laurowski:


markdown/3.x/en/documentation/server-side-rendering.md
```diff
@@ -57,9 +57,9 @@ Define the `ff-record` template which `ff-record-list` will use to render new re
...
-Check the `ff-record-list` [documentation](/api/3.x/ff-record-list#tab=docs) for more details about its template.  
+```
+Check the `ff-record-list` [documentation](/api/3.x/ff-record-list#tab=docs) for more details about its template.
 
-After doing this, the only thing left to do is to trigger a `search` event with the same query used on the server side.
-You can achieve this either by using [EventAggregator](/api/3.x/core-event-aggregator) or by setting the `search-immediate` attribute on `ff-communication`.
-This will trigger `ff-record-list` to replace pre-rendered records with the actual ones. 
+After doing this, the only thing left to do is to dispatch response to the Web Components using `dispatchRaw`.
+See the [Result Dispatcher](/api/3.x/core-result-dispatcher) for more details regarding this utility.
+Dispatching response via `dispatchRaw` will trigger `ff-record-list` to replace pre-rendered records with the actual ones.
```
omikron-fp, line 65:
```suggestion
After doing this, the only thing left to do is to dispatch the response to Web Components using `dispatchRaw`.
See [Result Dispatcher](/api/3.x/core-result-dispatcher) for more details regarding this utility.
Dispatching the response via `dispatchRaw` will trigger `ff-record-list` to replace pre-rendered records with the actual ones. 
```

markdown/3.x/en/core-result-dispatcher.md
```diff
@@ -107,3 +107,18 @@ registered callback.
 </script>
 <script defer src="pathToFFWebComponents/dist/bundle.js"></script>
 ```
+
+### `dispatchRaw(response, topics)`
+___
+Manually dispatches a response to the Web Components.
+Topics argument is optional - skipping it causes the response will be dispatched as standard search result.
```
omikron-fp, line 114: Perhaps we could elaborate a bit on this.
For example why this method is useful, what its intended use case is and so on.
How does its behaviour differ when specifying `topics` compared to not specifying it?
If the response is not dispatched as a standard search result, what is it dispatched as instead?

markdown/3.x/en/core-result-dispatcher.md
```diff
@@ -107,3 +107,18 @@ registered callback.
...
+
+```html
+<script type="text/javascript">
+  document.addEventListener('WebComponentsReady', function () {
+      factfinder.communication.ResultDispatcher.dispatchRaw(responseToDispatch);
+  });
+</script>
+```
+**Note**: Make sure you put `dispatchRaw` call inside `WebComponentsReady` listener.
+That guarantees all attributes set for `ff-communication` (e.g. `version`) will be reflected as communication parameters.
```
omikron-fp, line 124:
```suggestion
**Note**: Make sure you put the `dispatchRaw` call inside a `WebComponentsReady` listener.
This guarantees all `ff-communication` attributes (e.g. `version`) will be reflected as communication parameters.
```

omikron-fp (changes requested):

markdown/3.x/en/documentation/server-side-rendering.md
```diff
@@ -35,7 +35,7 @@ Example libraries (the whole list of libraries can be found at [mustache.github.
 **Note**: You are not just limited to mustache. You can use any template engine your platform supports. The advantage of using mustache here is that you can use the same templates as in Web Components.  
 
 #### Sending the response
-The response to the browser contains the pre-rendered HTML.
+The response to the browser contains the pre-rendered HTML and raw response available to be used by Web Components.
```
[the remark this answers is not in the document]
a-laurowski: added more info

markdown/3.x/en/core-result-dispatcher.md
a-laurowski: I updated the description. Hope that this level of details makes things more clear with not confusing reader in the same time

markdown/3.x/en/core-result-dispatcher.md
```diff
@@ -107,3 +107,23 @@ registered callback.
...
+
+### `dispatchRaw(response, topics)`
+___
+Manually dispatches a **response** to the Web Components.
+This method is useful when there is a need to dispatch FACT-Finder response received in different way than by AJAX request, sent by Web Components.
+That will be the case especially when implementing [Server Side Rendering](/documentation/3.x/server-side-rendering).
+
+Argument **topics** is optional - skipping it causes the response will be dispatched to all default topics causing updates all the actively subscribing elements.
+With **topics** argument passed only a part of the response, related to the passed topic will be dispatched (e.g. using`suggest` will cause an update only on elements which subscribe to that topic - by default `ff-suggest-element`).
+**Note** It is possible to pass multiple topics in an array.
```
omikron-fp, line 119: Is the note intended to be inline? Using a new paragraph like the next note might be more of an eye-catcher.
Or a forced linebreak could be enough already.
```suggestion
**Note**: It is possible to pass multiple topics in an array.
```

markdown/3.x/en/documentation/server-side-rendering.md
```diff
@@ -35,7 +35,15 @@ Example libraries (the whole list of libraries can be found at [mustache.github.
...
+Most convenient way to make the response available to use by the Web Components is to store it as JSON in a variable.
```
omikron-fp, line 39:
```suggestion
The response to the browser contains the pre-rendered HTML and the raw response ready to be used by Web Components.
The most convenient way to make the response available to use by Web Components is to store it as JSON in a variable.
```

markdown/3.x/en/documentation/server-side-rendering.md
```diff
@@ -35,7 +35,15 @@ Example libraries (the whole list of libraries can be found at [mustache.github.
...
+```html
+<script>
+const rawResponse = JSON.parse(RESPONSE_TEXT);
+</script>
```
omikron-fp, line 43: Let's indent this to be consistent with other listings.
There is another listing that could be indented on line 50.
```suggestion
<script>
    const rawResponse = JSON.parse(RESPONSE_TEXT);
</script> 
```

markdown/3.x/en/documentation/server-side-rendering.md
```diff
@@ -35,7 +35,15 @@ Example libraries (the whole list of libraries can be found at [mustache.github.
...
 #### Sending the response
-The response to the browser contains the pre-rendered HTML.
+The response to the browser contains the pre-rendered HTML and raw response available to be used by Web Components.
+Most convenient way to make the response available to use by the Web Components is to store it as JSON in a variable.
+```html
+<script>
+const rawResponse = JSON.parse(RESPONSE_TEXT);
+</script> 
+```
+**Note**: `RESPONSE_TEXT` in example above represents response text received from FACT-Finder. Passing it to variable depends on the used framework.
```
omikron-fp, line 45:
```suggestion
**Note**: `RESPONSE_TEXT` in the example above represents the response text received from FACT-Finder. The way it is passed to a variable depends on the used framework.
```

markdown/3.x/en/documentation/server-side-rendering.md
```diff
@@ -57,9 +65,16 @@ Define the `ff-record` template which `ff-record-list` will use to render new re
     <template data-role="record">
         <!--put `ff-record` element template here  --->
     </template>
```
omikron-fp, line 67: While we are at it.
```suggestion
<template data-role="record">
    <!--put `ff-record` element template here  --->
</template>
```

markdown/4.x/en/core-result-dispatcher.md
```diff
@@ -107,3 +107,23 @@ registered callback.
 </script>
 <script defer src="pathToFFWebComponents/dist/bundle.js"></script>
 ```
+
+### `dispatchRaw(response, topics)`
+___
+Manually dispatches a **response** to the Web Components.
+This method is useful when there is a need to dispatch FACT-Finder response received in different way than by AJAX request, sent by Web Components.
+That will be the case especially when implementing [Server Side Rendering](/documentation/3.x/server-side-rendering).
```
omikron-fp, line 115:
```suggestion
That will be the case especially when implementing [Server Side Rendering](/documentation/4.x/server-side-rendering).
```

omikron-fp (changes requested):
This is a good balance of detail and brevity I think. Just some grammar things.

I only inspected the `3.x` files assuming they are identical to `4.x`.
`4.x` contains only one reference error it seems.

markdown/3.x/en/core-result-dispatcher.md
```diff
@@ -107,3 +107,23 @@ registered callback.
...
 ```
+
+### `dispatchRaw(response, topics)`
+___
+Manually dispatches a **response** to the Web Components.
+This method is useful when there is a need to dispatch FACT-Finder response received in different way than by AJAX request, sent by Web Components.
+That will be the case especially when implementing [Server Side Rendering](/documentation/3.x/server-side-rendering).
+
+Argument **topics** is optional - skipping it causes the response will be dispatched to all default topics causing updates all the actively subscribing elements.
+With **topics** argument passed only a part of the response, related to the passed topic will be dispatched (e.g. using`suggest` will cause an update only on elements which subscribe to that topic - by default `ff-suggest-element`).
```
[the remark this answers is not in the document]
a-laurowski: Its misstype. That should be the `ff-suggest` element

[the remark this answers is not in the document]
omikron-fp: I see what you mean. In this case I think the formatting in `removeCallback` is wrong. At the top under `subscribe`, monospace is used to refer to the parameters.
Allow me to adjust this quickly.

omikron-fp (approved):

PR #271, opened 2020-09-17
FFWEB-1802: Document 4.x ff-suggest behaviour

a-laurowski:


markdown/4.x/en/ff-suggest.md
```diff
@@ -199,3 +199,18 @@ See the following example:
...
 ```
+
+## On-Click redirect
+The expected behavior after clicking on `ff-suggest-item` element is a redirection to the clicked record's page.
+`ff-suggest` support this functionality if only `deeplink` field is available in clicked item attributes data.
+The `deeplink` field is populated using the data from data feed export column with a field role with the same name.
+
+**Note**: `ff-suggest` is redirecting both absolute and relative URLs.
+
+If no `deeplink` is provided for anitem which is of `productName` type then `ff-suggest` will try to obtain its value by emitting additional request.
```
omikron-fp, line 210:
```suggestion
If no `deeplink` is provided for an item of type `productName`, then `ff-suggest` will try to obtain its value by emitting an additional request.
```

markdown/4.x/en/ff-suggest.md
```diff
@@ -199,3 +199,18 @@ See the following example:
...
+Depending on used FACT-Finder version it would be a search - for versions lower than `7.3`, or record API call - for newer versions.
+Then`ff-suggest` will try to redirect user once more, this time using the data received from the response to the mentioned request.
```
omikron-fp, line 212: Parentheses might be a little clearer about their reference than dashes.
```suggestion
Depending on the used FACT-Finder version it will be a search request (versions lower than `7.3`), or a records API call (versions `7.3` and up).
Then `ff-suggest` will try to redirect the user once more, this time using the data received from the response to the mentioned request.
```

markdown/4.x/en/ff-suggest.md
```diff
@@ -199,3 +199,18 @@ See the following example:
...
+
+If clicked item is not of `productName` type, a standard search request will be issued to FACT-Finder with additional parameter `queryFromSuggest` which is unique to `ff-suggest` and might be used to identify search request issued by this element.
```
omikron-fp, line 214:
```suggestion
If the clicked item is not of type `productName`, a standard search request will be issued to FACT-Finder with an additional parameter `queryFromSuggest` which is unique to `ff-suggest` and may be used to identify search requests issued by this element.
```

markdown/4.x/en/ff-suggest.md
```diff
@@ -199,3 +199,18 @@ See the following example:
...
+
+**Note**: Setting `ffPreventDefault` on a given item before clicking it will prevent the redirection.
```
omikron-fp, line 216: The value `ffPreventDefault` is set to is important I think. We should mention it explicitly.
A tiny example would make it even easier for users I think.

omikron-fp (changes requested):

markdown/4.x/en/ff-suggest.md
```diff
@@ -199,3 +199,18 @@ See the following example:
     });
 </script>
 ```
+
+## On-Click redirect
+The expected behavior after clicking on `ff-suggest-item` element is a redirection to the clicked record's page.
+`ff-suggest` support this functionality if only `deeplink` field is available in clicked item attributes data.
+The `deeplink` field is populated using the data from data feed export column with a field role with the same name.
```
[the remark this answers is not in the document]
a-laurowski: You mean this line ?

> The `deeplink` field is populated with the data from the data feed export column configured with the `Deeplink` field role.

Yes it is correct

markdown/4.x/en/ff-suggest.md
a-laurowski: What do you think about only pointing to the paragraph above "Overriding default click action"? There's a good example of how to set `ffPreventDefault`

omikron-fp: Yes, that sounds reasonable as long as it is clear what `ffPreventDefault` has to be set to.

a-laurowski: so i added the info about the expected value of `ffPreventDefault` and point to section "Overriding default click action" for more info and example usage

omikron-fp (approved):

markdown/4.x/en/ff-suggest.md
```diff
@@ -199,3 +199,20 @@ See the following example:
...
+
+**Note**: `ff-suggest` is redirecting both absolute and relative URLs.
+
+If no `deeplink` is provided for an item of type `productName`, then `ff-suggest` will try to obtain its value by emitting an additional request.
+Depending on the used FACT-Finder version it will be a search request (versions lower than `7.3`), or a records API call (versions `7.3` and up).
+Then `ff-suggest` will try to redirect the user once more, this time using the data received from the response to the mentioned request.
+
+If the clicked item is not of type `productName`, a standard search request will be issued to FACT-Finder with an additional parameter `queryFromSuggest` which is unique to `ff-suggest` and may be used to identify search requests issued by this element.
+
+To prevent the built-in behaviour, please set `ffPreventDefault` to `true` on a specific suggest item before the default event listener reacts to the `click` event.
```
omikron-fp, line 216: Note that we have `behaviour` here and `behavior` further up in the file.
Perhaps you want to unify this before merging.

a-laurowski: sure, will unify that

a-laurowski: @omikron-fp  Changed. I found other occurrences of `behaviour` in the text. Should i change them too, since i'm on that?

omikron-fp: If you like, sure 👍

a-laurowski: done

PR #274, opened 2020-11-05
FFWEB-1849: `idType` attribtue for ff-similar-products

a-laurowski:


fmarangi (approved):

markdown/3.x/en/api/ff-similar-products.api.md
```diff
@@ -4,6 +4,7 @@ ___
 | Name | Description |
 | ---- | ----------- |
 | **record-id**&nbsp;(Number) (default: empty) | The record ID of the product, for which similar products are returned. This also works for multiple products. In this case the record IDs have to be separated with `,`. |
+| **id-type**&nbsp;(String) **Options**: &nbsp;"id", &nbsp;"productNumber" (default: undefined) | The type of identifier for which the corresponding value should be passed in the **record-id**. This attribute is mandatory when using FACT-Finder NG. |
```
omikron-fp, line 7: Honestly, I don't understand what this is trying to tell me.

Considering the FF API documentation
> Use this parameter to determine the type of id passed to the method.

Perhaps we can make something short and simple, too?

```suggestion
| **id-type**&nbsp;(String) **Options**: &nbsp;"id", &nbsp;"productNumber" (default: undefined) | This attribute determines the type of `record-id`. It is mandatory when using FACT-Finder NG. |
```

omikron-fp (changes requested):

markdown/3.x/en/api/ff-similar-products.api.md
a-laurowski: Yes i had difficulties to to put it somehow neatly, this version is much better. Changed in v4 too

omikron-fp:
Shall we introduce the (required/conditionally required) pattern like in `ff-communication`?

a-laurowski:
> Shall we introduce the (required/conditionally required) pattern like in `ff-communication`?

sure, we can. 
1. Do we update only v4? (like ff-communication)
2. Should we use"required" mark only with attributes which will not prevent element from working but will eventually break the execution and emit warning/error (like api,version, channel and rest from ff-communication), or include also these, which while not being set, will make element will not work at all (like record-id from this element)

omikron-fp:
Only applying it to `4.x` should be fine for now.
I think it would be best to mark all attributes as _required_ that are needed for proper functioning. So here it would be `record-id` and `id-type` I guess. Not sure though whether `id-type` isn't conditionally required.

a-laurowski:
> Only applying it to `4.x` should be fine for now.
> I think it would be best to mark all attributes as _required_ that are needed for proper functioning. So here it would be `record-id` and `id-type` I guess. Not sure though whether `id-type` isn't conditionally required.

ah yes, its mandatory only for NG. changed this and added record-id as required

omikron-fp (approved):

PR #277, opened 2020-11-16
FFWEB-1850 require version and api

omikron-fp:
Updates `4.0.0` migration guide and `ff-communication` API page.

### DIFF
--- a/markdown/4.x/en/api/ff-communication.api.md
+++ b/markdown/4.x/en/api/ff-communication.api.md
@@ -6,9 +6,9 @@ _(`*` - required, `†` - conditionally required)_
 
 | Name | Description |
 |----|-----------|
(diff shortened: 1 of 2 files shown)


markdown/4.x/en/api/ff-communication.api.md
```diff
@@ -1,11 +1,11 @@
 ## `ff-communication`
 ___
-### Properties
+### Properties (`*` - required)
```
KrzysztofKozlowskiOBSS, line 3: Shouldn't `url` and `channel` be marked as required too?

omikron-fp: I think so. I left them out here because they are somewhat unrelated nd there is a branch specifically for the `*`. Clarifying.

KrzysztofKozlowskiOBSS (approved):

markdown/4.x/en/api/ff-communication.api.md
omikron-fp: Generically added the "required" markers and rebuilt the history.

KrzysztofKozlowskiOBSS (approved):

PR #283, opened 2021-02-02
FFWEB-1145 Added info about sid required length

a-laurowski:


markdown/4.x/en/api/ff-communication.api.md
```diff
@@ -29,7 +29,7 @@ _(`*` - required, `†` - conditionally required)_
 | **use-aso**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "true") | Allows activating/deactivating of automated search optimization. true = the search result is automatically optimized. false = the search result is not optimized. |
 | **use-browser-history**&nbsp;(String) **Options**: &nbsp;"true", &nbsp;"false" (default: "true")| If set to true, the search history is pushed to the browser history, even without using URL parameter. |
 | **user-id**&nbsp;(String) (default: '') | As soon as this attribute is set, a Login tracking request is sent to FACT-Finder. |
-| **sid**&nbsp;(String) **any** | If set, the value provided in this property is used in every request as FACT-Finder session id parameter (sid). |
+| **sid**&nbsp;(String) **any** | A session identifier parameter used in every request sent to FACT-Finder. We recommend no to set this attribute manually as it is generated automatically by Web Components. If for some case you have to overwrite it, please keep in mind that it must be **exactly 30 character length**. |
```
omikron-fp, line 32: I think we can get rid of the `any`. Not sure what it is supposed to represent.
```suggestion
| **sid**&nbsp;(String) | A session identifier used in every request sent to FACT-Finder. We recommend not to set this attribute manually as it is generated automatically by Web Components. If for some reason you have to overwrite it, please note that it must be **exactly 30 characters long**. |
```

omikron-fp (changes requested):

markdown/4.x/en/api/ff-communication.api.md
a-laurowski: I think **any**  was placed here to indicate that there was no additional requirements of how the sid value should look like. Which is not true. Shall i remove it from the previous versions  too?

omikron-fp: If you like.

omikron-fp (approved):

PR #285, opened 2021-03-02
FFWEB-1908 upgrade routing code snippet to work with 4.x

KrzysztofKozlowskiOBSS:


omikron-fp (approved):
Does this also apply to `3.x`? If so, let's update that part of the docs, too.

KrzysztofKozlowskiOBSS:
> Does this also apply to `3.x`? If so, let's update that part of the docs, too.

It doesn't. I think our encoding related changes caused some issues with the old approach.

PR #287, opened 2021-05-25
One Touch Slider

omikron-fp:
Introduces a section in ASN about One Touch Slider.
Adds special styles in order to center the new images.

Removed "Styling slider groups" as it didn't reflect current behaviour. An empty `ff-slider-control` element renders a broken default template.


a-laurowski (approved):
good idea with the one touch slider GIF  👍

PR #296, opened 2024-05-25
rebranding, dependencies update

KrzysztofKozlowskiOBSS:
image
image
image


markdown/1.x/en/documentation/communication.md
```diff
@@ -50,10 +50,10 @@ now to be made to the shop system now. Change the URL settings in `ff-communicat
...
 
 
-#### 2. Forward the HTTP Request to FACT-Finder
+#### 2. Forward the HTTP Request to FactFinder
 
 ---
-The request made by FACT-Finder Web Components will look something like: 
+The request made by FactFinder Web Components will look something like: 
 
     `http://domain.com/FACT-Finder/Search.ff?query=shoe&channel=shoeshop_en`.
```
KrzysztofKozlowskiOBSS, line 58: I intentionally skipped replacing `FACT-Finder` in the URLs. Please confirm if that's expected.

src/documentation-app.js
```diff
@@ -36,7 +36,7 @@ class DocumentationApp extends ReduxMixin(PolymerElement) {
         --app-primary-color: #3c3c3b;;
         --app-secondary-color: white;
       
-        --button-active-background-color: #1b4385;
+        --button-active-background-color: #a175ea;
```
KrzysztofKozlowskiOBSS, line 39: Zrzut ekranu 2024-05-25 o 23 05 34
This color was not specified in the presentation, so I just used color picker to extract it

src/styles/sidebar-styles.js
```diff
@@ -57,8 +57,8 @@ $_documentContainer.innerHTML = `<dom-module id="sidebar-styles">
             }
 
             .panel-menus > iron-selector a.iron-selected {
-                border-left: 6px solid rgb(27, 67, 133);
-                color: rgb(27, 67, 133);
+                border-left: 6px solid #230871;
+                color: #230871;
```
KrzysztofKozlowskiOBSS, line 61: Zrzut ekranu 2024-05-25 o 23 09 48

these panels weren't mentioned in the design presentation, but I think they look better with the new color, too. Let me know if it should stay like that.

src/views/home-view.js
```diff
@@ -90,7 +90,7 @@ class HomeView extends ViewMixin(ReduxMixin(PolymerElement)) {
     }
 
     button:hover {
-        background-color: #007bbf;
+        background-color: #230871;
```
KrzysztofKozlowskiOBSS, line 93: image
this button previously had a hover effect. I temporarily set the hover color to the same as regular. Let me know if we can turn off the hover effect completely. Or maybe this button should keep its old colors?

markdown/1.x/en/documentation/communication.md
omikron-fp: Yes, I think this was the right thing to do.

src/styles/sidebar-styles.js
omikron-fp: Looks good! Let's keep it.

src/views/home-view.js
omikron-fp: Good call on this one! I think it's only consistent to also apply the new colours to this button.
We can do without the hover effect.

omikron-fp (approved):

src/views/home-view.js
KrzysztofKozlowskiOBSS: hover effect removed

